PR #15, opened 2018-10-17
DIT-228 Add Case Summary

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/connector/CasesConnector.scala
```diff
@@ -30,6 +30,11 @@ import scala.concurrent.Future
 @Singleton
 class CasesConnector @Inject()(configuration: AppConfig, client: HttpClient) {
 
+  def getOne(reference: String)(implicit hc: HeaderCarrier): Future[Option[Case]] = {
+    val url = configuration.bindingTariffClassificationUrl + "/cases/" + reference
```
pllaurado: I don't know if we are going to call more services, but I would prefer to create a common structure for all of them
myService:
    protocol: https
    domain: myService.qa.gov.uk
    port: 1111
    timeout: 5s
    ...

This is the minimum I would set on a config file. we could set as well the endpoints but this point could be discussed

app/uk/gov/hmrc/tariffclassificationfrontend/connector/CasesConnector.scala
```diff
@@ -30,6 +30,11 @@ import scala.concurrent.Future
...
+    client.GET[Option[Case]](url)
+  }
+
   def getGatewayCases(implicit hc: HeaderCarrier): Future[Seq[Case]] = {
     val url = configuration.bindingTariffClassificationUrl + "/cases?queue_id=none&assignee_id=none&sort-by=elapsed-days"
```
pllaurado: To improve: Maybe I miss this one, but use a URIBuilder to handle this lines would be great

app/uk/gov/hmrc/tariffclassificationfrontend/connector/CasesConnector.scala
ed0906: Have a look at the `microservice.service` block in `application.conf`. This gets picked up by `AppConfig` (see line 38).

app/uk/gov/hmrc/tariffclassificationfrontend/connector/CasesConnector.scala
ed0906: Potentially. @gokyo Might now if theres a Scala equivalent.

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/CasesController.scala
```diff
@@ -32,7 +32,7 @@ class CasesController @Inject()(casesService: CasesService,
                                 implicit val appConfig: AppConfig) extends FrontendController with I18nSupport {
 
   def gateway: Action[AnyContent] = Action.async { implicit request =>
-    casesService.getAllCases.map { cases =>
+    casesService.getGatewayCases.map { cases =>
```
pllaurado: Do we need one function for each queue?

app/uk/gov/hmrc/tariffclassificationfrontend/models/Application.scala
```diff
@@ -24,8 +24,24 @@ sealed trait Application {
   def asBTI(): BTIApplication = {
     this.asInstanceOf[BTIApplication]
   }
+
+  def asLiabilityOrder(): LiabilityOrder = {
+    this.asInstanceOf[LiabilityOrder]
+  }
+
+  def isBTI(): Boolean = {
```
pllaurado: Just a thought. Was the type created to solve this type of problems?

ed0906: For some reason the `type` field is not directly accessible in the templates :s It needs formatting anyway so its useful to have the `getType` method so we're not showing stuff like "LIABILITY_ORDER" in the UI

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/CasesController.scala
ed0906: I'm thinking we'll eventually have a `Queue` class which we can just pass into a `casesService.getCases(queue)` method. We can refactor on the next queue frontend task :)

app/uk/gov/hmrc/tariffclassificationfrontend/models/Application.scala
ed0906: Theres also going to be some logic around what type of application it is (e.g. showing them in different views) so I think its quite nice to have an `isBTI` `isLiability` method rather than doing `instanceOf` or `type == "BTI"` everywhere.  I'm open to alternatives

ed0906: Also, to be honest I think the type field is for deserialisation, it shouldn't even be in the model really as we can use the inheritance structure to work it out

app/uk/gov/hmrc/tariffclassificationfrontend/connector/CasesConnector.scala
```diff
@@ -30,6 +30,11 @@ import scala.concurrent.Future
 @Singleton
 class CasesConnector @Inject()(configuration: AppConfig, client: HttpClient) {
```
gokyo: I would write for now this:
```
  def getOne(reference: String)(implicit hc: HeaderCarrier): Future[Option[Case]] = {
    val url = s"${configuration.bindingTariffClassificationUrl}/cases/$reference"
    client.GET[Option[Case]](url)
  }

  def getGatewayCases(implicit hc: HeaderCarrier): Future[Seq[Case]] = {
    val url = s"${configuration.bindingTariffClassificationUrl}/cases?queue_id=none&assignee_id=none&sort-by=elapsed-days"
    client.GET[Seq[Case]](url)
  }
```

ed0906: Will do, It's probably the Java developer in me that prefers the `+`

test/unit/uk/gov/hmrc/tariffclassificationfrontend/connector/CasesConnectorSpec.scala
```diff
@@ -43,12 +42,10 @@ class CasesConnectorSpec extends UnitSpec with WiremockTestServer with MockitoSu
...
   "Connector" should {
 
-    val url = "/cases?queue_id=none&assignee_id=none&sort-by=elapsed-days"
-
-    "get empty cases" in {
+    "get empty gateway cases" in {
       given(configuration.bindingTariffClassificationUrl).willReturn("http://localhost:20001")
 
-      stubFor(get(urlEqualTo(url))
+      stubFor(get(urlEqualTo("/cases?queue_id=none&assignee_id=none&sort-by=elapsed-days"))
```
gokyo: save those urls as `val`, they are in fact reused in other tests

test/unit/uk/gov/hmrc/tariffclassificationfrontend/utils/DatesSpec.scala
```diff
@@ -14,12 +14,11 @@
  * limitations under the License.
  */
 
-package unit.uk.gov.hmrc.tariffclassificationfrontend.utils
+package uk.gov.hmrc.tariffclassificationfrontend.utils
```
gokyo: revert

test/unit/uk/gov/hmrc/tariffclassificationfrontend/utils/CasePayloads.scala
```diff
@@ -14,24 +14,15 @@
  * limitations under the License.
  */
 
-package unit.uk.gov.hmrc.tariffclassificationfrontend.connector
-
-
-import java.time.ZonedDateTime
+package uk.gov.hmrc.tariffclassificationfrontend.utils
```
gokyo: revert

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/CaseController.scala
```diff
@@ -0,0 +1,45 @@
...
+import play.api.mvc._
+import uk.gov.hmrc.play.bootstrap.controller.FrontendController
+import uk.gov.hmrc.tariffclassificationfrontend.config.AppConfig
+import uk.gov.hmrc.tariffclassificationfrontend.service.CasesService
+import uk.gov.hmrc.tariffclassificationfrontend.views
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+@Singleton
+class CaseController @Inject()(casesService: CasesService,
```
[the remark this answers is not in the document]
ed0906: I'm going to keep it as it as I know it's only going to need splitting out again in the next week or so

test/unit/uk/gov/hmrc/tariffclassificationfrontend/utils/DatesSpec.scala
ed0906: Seems a bit weird though, the test should be in the same package as the thing it's testing. We have the ability to create different test source sets though

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/CaseController.scala
```diff
@@ -0,0 +1,45 @@
...
+                               val messagesApi: MessagesApi,
+                               implicit val appConfig: AppConfig) extends FrontendController with I18nSupport {
+
+  def summary(reference: String): Action[AnyContent] = Action.async { implicit request =>
+    casesService.getOne(reference)
```
gokyo: You can make it more concise:
```
  casesService.getOne(reference).map { response => 
    if (response.isEmpty) Ok(views.html.case_not_found(reference))
    else Ok(views.html.case_summary(response.get))
   }
```

app/uk/gov/hmrc/tariffclassificationfrontend/models/Application.scala
```diff
@@ -24,8 +24,24 @@ sealed trait Application {
...
+  }
+
+  def isLiabilityOrder(): Boolean = {
+    this.isInstanceOf[LiabilityOrder]
+  }
+
   def getType() : String = {
-    `type`
+    `type` match {
+      case "BTI" => "BTI"
```
gokyo: Is "BTI Application" more appropriate ?

test/unit/uk/gov/hmrc/tariffclassificationfrontend/utils/DatesSpec.scala
gokyo: It is a personal preference.
But I like the idea of having all tests inside a single folder (`test`)

app/uk/gov/hmrc/tariffclassificationfrontend/models/Application.scala
ed0906: At the moment the requirement is just to show the type as "BTI" or "Liability Order". Potentially a question for Joe

test/unit/uk/gov/hmrc/tariffclassificationfrontend/utils/DatesSpec.scala
ed0906: image
This is how it looks now

PR #33, opened 2018-10-29
DIT-289 Minor fixes to rulings. Should show Edit button when no ruling available

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/models/Attachment.scala
```diff
@@ -38,7 +38,11 @@ case class Attachment
     }
   }
   def name: Option[String] = {
-    val path = new URL(url).getPath
-    if(path.isEmpty) None else Option(FilenameUtils.getName(path))
+    try {
+      val path = new URL(url).getPath
+      if(path.isEmpty) None else Option(FilenameUtils.getName(path))
+    } catch {
+      case _: Exception => None
```
pllaurado: why do we want to mask exceptions in this way?

PR #53, opened 2018-12-07
DIT-596: integration test

gokyo:


gokyo:
@ed0906 
Sorry, I saw your question only now.
The tests are similar, but:
- the unit test verifies the behaviour of the filter using the AppConfig object
- the integration tests verifies the behaviour of the filter using the Play configuration

PR #59, opened 2018-12-14
Dit 339 - Send the applicant an email when a case is marked as complete

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/connector/EmailConnector.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+@Singleton
+class EmailConnector @Inject()(configuration: AppConfig, client: HttpClient) {
+
+  def send[E >: Email[_]](e: E)(implicit hc: HeaderCarrier, writes: Writes[E]): Future[Unit] = {
+    val url = s"${configuration.emailUrl}/hmrc/email"
+    client.POST(url = url, body = e).map(_ => Unit)
```
gokyo: Is `.map(_ => Unit)` needed?

app/uk/gov/hmrc/tariffclassificationfrontend/connector/EmailConnector.scala
```diff
@@ -0,0 +1,38 @@
...
+import uk.gov.hmrc.tariffclassificationfrontend.config.AppConfig
+import uk.gov.hmrc.tariffclassificationfrontend.models.Email
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+@Singleton
+class EmailConnector @Inject()(configuration: AppConfig, client: HttpClient) {
+
+  def send[E >: Email[_]](e: E)(implicit hc: HeaderCarrier, writes: Writes[E]): Future[Unit] = {
```
gokyo: is it possible to use a type instead of `_` ?

app/uk/gov/hmrc/tariffclassificationfrontend/connector/EmailConnector.scala
ed0906: Yep, wont compile without

app/uk/gov/hmrc/tariffclassificationfrontend/connector/EmailConnector.scala
ed0906: for example?

ed0906: changed to Any

gokyo: With `Any` you are too permissive.
I will have a look and see for a better way

ed0906: I don’t think there is any reason to lock it down at this point

PR #65, opened 2019-01-08
Dit-659 Refactor Attachments to be owned by the Filestore

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/connector/FileStoreConnector.scala
```diff
@@ -0,0 +1,47 @@
...
+import uk.gov.hmrc.tariffclassificationfrontend.models.response.FileMetadata
+import uk.gov.hmrc.tariffclassificationfrontend.utils.JsonFormatters.fileMetaDataFormat
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+@Singleton
+class FileStoreConnector @Inject()(appConfig: AppConfig, http: HttpClient) {
+
+  def get(attachments: Seq[Attachment])(implicit headerCarrier: HeaderCarrier): Future[Seq[FileMetadata]] = {
```
pllaurado: I would maybe go for the minimum expression here and just make this connector know about ids.
**get (ids : Seq[String])
get (id   : String)**
It's more a preference than a requirement.  Just look more precise on the signature about what are you doing, as well as in the weird case that you could get attachments without ids ...

app/uk/gov/hmrc/tariffclassificationfrontend/connector/FileStoreConnector.scala
```diff
@@ -0,0 +1,47 @@
...
+import uk.gov.hmrc.tariffclassificationfrontend.config.AppConfig
+import uk.gov.hmrc.tariffclassificationfrontend.models._
+import uk.gov.hmrc.tariffclassificationfrontend.models.response.FileMetadata
+import uk.gov.hmrc.tariffclassificationfrontend.utils.JsonFormatters.fileMetaDataFormat
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+@Singleton
+class FileStoreConnector @Inject()(appConfig: AppConfig, http: HttpClient) {
```
gokyo: two methods to test and maintain for doing basically the same thing.... DRY principle

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/CaseController.scala
```diff
@@ -24,37 +24,52 @@ import uk.gov.hmrc.play.bootstrap.controller.FrontendController
...
 class CaseController @Inject()(authenticatedAction: AuthenticatedAction,
                                casesService: CasesService,
+                               fileStoreService: FileStoreService,
                                mapper: DecisionFormMapper,
                                val messagesApi: MessagesApi,
                                implicit val appConfig: AppConfig) extends FrontendController with I18nSupport {
 
   def summary(reference: String): Action[AnyContent] = authenticatedAction.async { implicit request =>
-    getCaseAndRenderView(reference, CaseDetailPage.SUMMARY, views.html.partials.case_summary(_))
+    getCaseAndRenderView(reference, CaseDetailPage.SUMMARY, c => successful(views.html.partials.case_summary(c)))
```
gokyo: if you revert this change, it will be more concise

app/uk/gov/hmrc/tariffclassificationfrontend/connector/FileStoreConnector.scala
ed0906: DRY principle is a **principle** and should be followed pragmatically. These endpoints are fulfilling two **different** purposes

app/uk/gov/hmrc/tariffclassificationfrontend/models/StoredAttachment.scala
```diff
@@ -0,0 +1,63 @@
...
+      case "image/gif" => true
+      case _ => false
+    }
+  }
+
+}
+
+object StoredAttachment {
+  def apply(attachment: Attachment, metadata: FileMetadata): StoredAttachment = {
+    if(attachment.id != metadata.id) {
```
gokyo: you can also use
```
require( attachment.id == metadata.id, "Cannot combine different attachments" )
```

ed0906: Will change

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/CaseController.scala
```diff
@@ -24,37 +24,52 @@ import uk.gov.hmrc.play.bootstrap.controller.FrontendController
...
   }
 
   def applicationDetails(reference: String): Action[AnyContent] = authenticatedAction.async { implicit request =>
-    getCaseAndRenderView(reference, CaseDetailPage.APPLICATION_DETAILS, views.html.partials.application_details(_))
+    getCaseAndRenderView(
+      reference,
+      CaseDetailPage.APPLICATION_DETAILS,
+      c => {
+        for {
```
gokyo: I think you should be able to define everything inside a single for comprehension:
```
         for {
          attachments <- fileStoreService.getAttachments(c)
          letter <- fileStoreService.getLetterOfAuthority(c)
          res <- views.html.partials.application_details(c, attachments, letter)
        } yield res
```

app/uk/gov/hmrc/tariffclassificationfrontend/service/FileStoreService.scala
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  def getLetterOfAuthority(c: Case)(implicit hc: HeaderCarrier): Future[Option[StoredAttachment]] = {
+    if(c.application.isBTI) {
+      c.application.asBTI.agent match {
+        case Some(agent: AgentDetails) =>
+          connector
+            .get(agent.letterOfAuthorisation)
+            .map(_.map(StoredAttachment(agent.letterOfAuthorisation, _)))
```
pllaurado: As on the other PR, this is a case where an agent exist and the letterOfAuth should be there, what we could do if the Option[] comes empty? 
Maybe this is a case where we just have NO letter... but it looks more like we have a Letter (attatchment) but we could not found it in the filestore.

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/CaseController.scala
ed0906: Will fix

app/uk/gov/hmrc/tariffclassificationfrontend/service/FileStoreService.scala
ed0906: I think from a UI perspective "having no letter" and "having a letter that doesn't exist" means the same thing (the operator has to contact the agent to ask for it)

app/uk/gov/hmrc/tariffclassificationfrontend/connector/FileStoreConnector.scala
ed0906: One endpoint always returns a `200`. The other returns a `200` or `404` (if it doesnt exist). They both have different roles and are important

app/uk/gov/hmrc/tariffclassificationfrontend/service/FileStoreService.scala
pllaurado: I completely agree but maybe a warning that a required file hasn't been found...

PR #76, opened 2019-01-18
Dit 372 add note to case

retroredge:


gokyo:
I know some of the code was not written by you... 
However, I would ask a small enhancement:
Instead of mocking with `any[String]`, I would use the real value of the case reference (you can achieve it by using `refEq` or by passing directly the value).
This would make the tests more strict and reliable

app/uk/gov/hmrc/tariffclassificationfrontend/service/EventsService.scala
```diff
@@ -32,4 +34,13 @@ class EventsService @Inject()(connector: BindingTariffClassificationConnector) {
     connector.findEvents(reference)
   }
 
+  def addNote(reference: String, note: String, operator: Operator, clock: Clock = Clock.systemDefaultZone())(implicit hc: HeaderCarrier): Future[Unit] = {
+    val event = NewEventRequest(Note(Some(note)), operator, ZonedDateTime.now(clock))
+    connector.createEvent(reference, event)
+      .recover {
+        case t: Throwable => Logger.error(s"Could not create Event for case [$reference] with payload [$event]", t)
```
gokyo: Is there a specific reason why you added this `recover` block?
If there is an error, I would prefer the future to fail.
We have the errors being logged in Kibana by the default error handler (defined in `application.conf`)

Typically we define `recover` blocks only for custom things ( for example if we want to add extra logging in the scheduled tasks). 
This seems something standard.

app/uk/gov/hmrc/tariffclassificationfrontend/service/EventsService.scala
```diff
@@ -32,4 +34,13 @@ class EventsService @Inject()(connector: BindingTariffClassificationConnector) {
...
+      }
+      .map(_ => ())
```
gokyo: Why are you mapping to `()` ?
I was wondering if it could be more correct to return here a `Future[Event]`  - as it comes from the connector.
Then in the controller you read the JSON response and display the fields you need in the front-end...

gokyo:
@retroredge 
The code looks alright, but I added some more comments.
We can discuss it on Monday. Would be good to know what @ed0906 and @pllaurado think about it

app/uk/gov/hmrc/tariffclassificationfrontend/service/EventsService.scala
retroredge: I was following the pattern from CasesService however I agree that the services should return the created item from the connector.

In this specific case we do a redirect after POST so the controller doesn't need the Event.

app/uk/gov/hmrc/tariffclassificationfrontend/service/EventsService.scala
retroredge: I was following the pattern from CasesService. I will come see you to discuss another way to code it

retroredge: Fixed

app/uk/gov/hmrc/tariffclassificationfrontend/service/EventsService.scala
retroredge: Fixed

app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
```diff
@@ -59,8 +59,8 @@ class BindingTariffClassificationConnector @Inject()(configuration: AppConfig, c
     client.PUT[Case, Case](url = url, body = c)
   }
 
-  def createEvent(c: Case, e: NewEventRequest)(implicit hc: HeaderCarrier): Future[Event] = {
-    val url = s"${configuration.bindingTariffClassificationUrl}/cases/${c.reference}/events"
+  def createEvent(reference: String, e: NewEventRequest)(implicit hc: HeaderCarrier): Future[Event] = {
```
ed0906: Personally, I prefer us to always pass full objects around e.g. Case rather than just references. It forces us to validate that the Case exists first and means our method signatures dont risk becoming `doSomething(string, string, string, int)` etc

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
```diff
@@ -37,7 +37,7 @@ class CasesService @Inject()(appConfig: AppConfig, auditService: AuditService,
   private def addEvent(original: Case, updated: Case, operator: Operator, comment: Option[String] = None)
                       (implicit hc: HeaderCarrier): Future[Unit] = {
     val event = NewEventRequest(CaseStatusChange(original.status, updated.status, comment), operator)
-    connector.createEvent(updated, event)
+    connector.createEvent(updated.reference, event)
```
ed0906: As mentioned above, in my opinion it was better as it was

app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
retroredge: I've changed the controller to check for case exists and passed the case down the stack

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
retroredge: Fixed

ed0906:
Looks good to me when conflicts are resolved

PR #83, opened 2019-01-29
DIT-372 ensure newline characters are rendered correctly

retroredge:


ed0906:
There is a progressive disclosure view test that will need a test adding in

app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/progressive_disclosure.scala.html
```diff
@@ -33,7 +33,7 @@
             </span>
         </summary>
         <div id="@{id}-body">
-            @body.get
+            @Html(HtmlFormat.escape(body.get).toString.replace("\n", "<br/>"))
```
ed0906: Can we add a test for this into ProgressiveDisclosureViewSpec please :)

retroredge: I've added a test for this

PR #90, opened 2019-02-01
DIT 394 - Add Commodity Code Search to the UI

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/forms/SearchFormData.scala
```diff
@@ -20,20 +20,22 @@ import play.api.data.Form
 import play.api.data.Forms._
 import uk.gov.hmrc.tariffclassificationfrontend.models.Search
 
-case class SearchFormData(traderName: String = "")
+case class SearchFormData(traderName: String = "", commodityCode: String = "")
```
ed0906: FYI I dont think forms can handle Options so it has to be an empty String... It gets filtered though in the Query String Binders

retroredge:
Look good to me

PR #91, opened 2019-02-04
BAU - Add ids and fix description behaviour

pllaurado:
BAU - Add ids and fix description behaviour


test/unit/uk/gov/hmrc/tariffclassificationfrontend/views/partials/AdvancedSearchResultsViewSpec.scala
```diff
@@ -89,6 +89,7 @@ class AdvancedSearchResultsViewSpec extends ViewSpec {
       // Then
       doc should containElementWithID("advanced_search_results-row-0-description")
       doc.getElementById("advanced_search_results-row-0-description") should containText("x" * 50 + "...")
+      doc.getElementById("advanced_search_results-row-0-description").text.size shouldBe 53
```
SteveSugden: Should the description not be exactly (i.e. equal) 50 x's plus '...' (i.e. "xxxxxx..xxxx..."

PR #94, opened 2019-02-06
DIT 398 - Add 'In Progress' filter

ed0906:
The following scenarios should apply:

`Front end Call` => `Back End Call`

`/search?include_in_progress=` => `/cases?status=COMPLETED&min_decision_end={current-time}&sort...`

`/search?include_in_progress=false` => `/cases?status=COMPLETED&min_decision_end={current-time}&sort...`

`/search?include_in_progress=true` => `/cases?sort...`

### DIFF
--- a/app/uk/gov/hmrc/tariffclassificationfrontend/controllers/SearchController.scala
+++ b/app/uk/gov/hmrc/tariffclassificationfrontend/controllers/SearchController.scala
@@ -47,12 +47,21 @@ class SearchController @Inject()(authenticatedAction: AuthenticatedAction,
           Future.successful(Results.Ok(html.advanced_search(formWithErrors, None)))
         },
         data => {
-          casesService.search(search, sort) map { results =>
+          val searchWithDefaults = applyDefaultFilters(search)
+          casesService.search(searchWithDefaults, sort) map { results =>
             Results.Ok(html.advanced_search(SearchForm.form.fill(data), Some(results)))
           }
         }
       )
 
+    }
+  }
+
+  private def applyDefaultFilters(search: Search): Search = {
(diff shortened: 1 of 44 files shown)


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/SearchController.scala
```diff
@@ -56,4 +57,12 @@ class SearchController @Inject()(authenticatedAction: AuthenticatedAction,
     }
   }
 
+  private def applyDefaultFilters(search: Search): Search = {
```
ed0906: Defaulting to live decisions only is logic specific to the "Advanced Search" view, so I stuck the logic here so we can potentially re-use Search in other places

app/uk/gov/hmrc/tariffclassificationfrontend/forms/SearchFormData.scala
```diff
@@ -19,26 +19,22 @@ package uk.gov.hmrc.tariffclassificationfrontend.forms
...
+  includeInProgress: Option[Boolean] = None
+)
 
 object SearchForm {
 
   val form = Form(
     mapping(
       "trader_name" -> optional(text.verifying(_ => true)),
-      "commodity_code" -> optional(text.verifying(emptyOr(numeric, minLength(2)): _*))
+      "commodity_code" -> optional(text.verifying(emptyOr(numeric, minLength(2)): _*)),
```
gokyo: you could add a constraint on the max length as well

ed0906: Done

PR #100, opened 2019-02-13
DIT-375 Reject case

pllaurado:


gokyo:
@pllaurado 
Can you add a scenario of your action in `AuditServiceTest` ?

gokyo:
@pllaurado 
Can you create a file `CasesService_RejectCaseSpec.scala` and add similar tests as done for the other actions?

app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/reject_suspend_section.scala.html
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.tariffclassificationfrontend.models.{Case, CaseStatus}
+@(c: Case)
+
+@if(c.status == CaseStatus.OPEN) {
+    <div class="mt-1">
```
ed0906: Imagine its worth sticking an ID on this so the logic can be tested?

ed0906:
I think there's some `ViewSpec` missing

ed0906:
Looks fine to me, I'd pull in master first though given we're all working in the same area

PR #107, opened 2019-02-14
DIT-376 case suspended

pllaurado:


gokyo:
@pllaurado 
Could you please add also an integration test?

PR #113, opened 2019-02-18
DIT 674 - Extended Use Flag

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/ExtendedUseCaseController.scala
```diff
@@ -0,0 +1,79 @@
...
+
+          }
+        )
+      }
+    )
+
+  }
+
+  private def statusHasChanged(c: Case, status: Boolean): Boolean = {
+    !c.decision.map(_.applicationForExtendedUse).contains(status)
```
pllaurado: contains?

ed0906: I'm not sure what your question is @pllaurado

pllaurado: I'm just curious  why you are comparing two booleans with contains. Not wrong but just estrange.  Maybe I would expect something like a filter and a isdefined.

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/ExtendedUseCaseController.scala
```diff
@@ -0,0 +1,79 @@
...
+          reference,
+          c => successful(views.html.change_extended_use_status(c, errors))
+        )
+      },
+      (status: Boolean) => {
+        getCaseAndRespond(
+          reference,
+          c => {
+            if(statusHasChanged(c, status)) {
+              caseService.updateExtendedUseStatus(c, status, request.operator).flatMap { c =>
```
pllaurado: The redirection is the same even the status has not change. In this case I would move the status change logic/check to the service.

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/ExtendedUseCaseController.scala
ed0906: It's the same thing really

PR #114, opened 2019-02-18
Dit 677

SteveSugden:


app/uk/gov/hmrc/tariffclassificationfrontend/models/Decision.scala
```diff
@@ -29,5 +31,6 @@ case class Decision
   methodExclusion: Option[String] = None,
   methodCommercialDenomination: Option[String] = None,
   appeal: Option[Appeal] = None,
-  review: Option[Review] = None
+  review: Option[Review] = None,
+  cancelReason: Option[CancelReason] = None
```
ed0906: Just a thought - I wonder if we should have done a `Cancellation` object which includes a `CancelationReason` and an `applicationForExtendedUse` flag

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
```diff
@@ -153,20 +154,24 @@ class CasesService @Inject()(appConfig: AppConfig,
...
+        cancelReason = Some(reason)
+      )
     val caseUpdating = original.copy(status = CaseStatus.CANCELLED, decision = Some(decisionUpdating))
 
     for {
       // Update the case
       updated: Case <- connector.updateCase(caseUpdating)
       // Create the event
-      _ <- addStatusChangeEvent(original, updated, operator)
+      _ <- addStatusChangeEvent(original, updated, operator, comment = Some(CancelReason.format(reason)))
```
ed0906: It's worth checking with Joe if there is a need to ever add an extra Note when cancelling a case. We'd need to keep the comment field free for that if we do

ed0906: Or we can concat them both I guess

app/uk/gov/hmrc/tariffclassificationfrontend/views/forms/components/input_radiogroup.scala.html
```diff
@@ -22,13 +22,14 @@
         inputClass: Option[String] = None,
         hint: Option[String] = None,
         labelClass: Option[String] = None,
+        heading:String = "Review status",
```
ed0906: Do you mind if we pull the heading out of this class. Seems a bit weird to have a default value of "Review Status" on a fairly generic component. 
Admittedly I think I made this so I should have spotted it originally...

test/unit/uk/gov/hmrc/tariffclassificationfrontend/controllers/CancelRulingControllerSpec.scala
```diff
@@ -103,16 +103,30 @@ class CancelRulingControllerSpec extends WordSpec with Matchers with UnitSpec
...
-      val result: Result = await(controller.confirmCancelRuling("reference")(newFakePOSTRequestWithCSRF(app)))
+      val result: Result = await(controller.confirmCancelRuling("reference")(newFakePOSTRequestWithCSRF(app).withFormUrlEncodedBody("reason" -> "ANNULLED")))
 
       status(result) shouldBe Status.OK
       contentTypeOf(result) shouldBe Some(MimeTypes.HTML)
       charsetOf(result) shouldBe Some("utf-8")
       bodyOf(result) should include("This case has been cancelled")
     }
 
+    "display required field when failing to submit reason" in {
```
ed0906: I'd maybe suggest this kind of test goes in the ViewSpec for that component e.g.` InputRadioGroupViewSpec`

app/uk/gov/hmrc/tariffclassificationfrontend/views/forms/components/input_radiogroup.scala.html
SteveSugden: Do you mean pull it out completely (so the component doesn't render a heading) or just remove the default value?  I agree that a "review" specific default is daft

app/uk/gov/hmrc/tariffclassificationfrontend/views/cancel_ruling.scala.html
```diff
@@ -47,11 +45,7 @@ <h3 class="heading-large mt-0">Cancel the ruling</h3>
                 <li><a href="@uk.gov.hmrc.tariffclassificationfrontend.controllers.routes.AttachmentsController.attachmentsDetails(c.reference)">attach</a>  any relevant documentation to the case</li>
             </ul>
 
-            @form(action = uk.gov.hmrc.tariffclassificationfrontend.controllers.routes.CancelRulingController.confirmCancelRuling(c.reference), 'class -> "form") {
-                @CSRF.formField
-                <button class="button" id="cancel-ruling-button" type="submit">Cancel ruling</button>
-                <a class="secondary-button" id="cancel-button" href="@uk.gov.hmrc.tariffclassificationfrontend.controllers.routes.CaseController.rulingDetails(c.reference)">Cancel</a>
-            }
+            @cancel_ruling_form(c, form("reason"))
```
pllaurado: I think this cancel_ruling exists because the form to cancel then separate it into another class (is not bad), how granular we want to be?? My opinion is always the clearest and easy to maintain option. But I think that having a form object and parsing just a field of this form into another component is a good reason to not have a separate component. What do you think?

SteveSugden: No strong opinion either way.  As the cancel form component is not really re-usable its contents could have been in-lined.   My defence is that I was following patterns seen elsewhere.

app/uk/gov/hmrc/tariffclassificationfrontend/views/forms/components/input_radiogroup.scala.html
pllaurado: I agree to move the heading to the top, but default it to Review status sounds the object will be less generic.

ed0906: Pull it out completely, I think there would be times we would want radio groups but not want the heading

app/uk/gov/hmrc/tariffclassificationfrontend/views/cancel_ruling.scala.html
ed0906: I dont see any harm pulling it out. although I agree its not particularly reusable

PR #117, opened 2019-02-18
BAU - Pull out common logic into trait

ed0906:
The Appeal, Review & Extended Use Status Change controllers have very similar logic. We could pull this out into a trait. Let me know your thoughts

### DIFF
--- a/app/uk/gov/hmrc/tariffclassificationfrontend/controllers/StatusChangeAction.scala
+++ b/app/uk/gov/hmrc/tariffclassificationfrontend/controllers/StatusChangeAction.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2019 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


ed0906:
There's other options e.g.
- Instead of using inheritance, define a generic `updateStatus` method where we pass in various args required in the trait above
-  Leave it as it is

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/StatusChangeAction.scala
```diff
@@ -0,0 +1,65 @@
...
+        getCaseAndRenderView(
+          reference,
+          c => successful(chooseStatusView(c, errors))
+        )
+      },
+      (status: T) => {
+        getCaseAndRespond(
+          reference,
+          c => {
+            if(statusHasChanged(c, status)) {
```
pllaurado: not about the refactor. But maybe move this logic into the service?

ed0906: Potentially I guess the question is: Is checking if a status has changed business logic, or view layer?

I don't have a strong opinion either way

gokyo:
I haven't looked at the code properly, but extracting the common code in a trait makes sense to me 👍

PR #119, opened 2019-02-18
DIT-677: Classification Officers can pick a reason (code) when cancelling a case

SteveSugden:


app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
```diff
@@ -166,20 +167,24 @@ class CasesService @Inject()(appConfig: AppConfig,
...
+        cancellation = Some(Cancellation(reason = reason))
+      )
     val caseUpdating = original.copy(status = CaseStatus.CANCELLED, decision = Some(decisionUpdating))
 
     for {
       // Update the case
       updated: Case <- connector.updateCase(caseUpdating)
       // Create the event
-      _ <- addStatusChangeEvent(original, updated, operator)
+      _ <- addStatusChangeEvent(original, updated, operator, comment = Some(CancelReason.format(reason)))
```
ed0906: Similar point to previously. I think we should check with Joe if there is a requirement coming for us to add a Comment to a Cancellation. 
Otherwise we might need to stick this somewhere else, or concat it

It's better we find out now then have to do data migration in the future

ed0906: I think we also discussed there was a future reporting requirement where we will need to report on how many cases have cancellations with code `x`

app/uk/gov/hmrc/tariffclassificationfrontend/audit/AuditService.scala
```diff
@@ -89,6 +89,7 @@ class AuditService @Inject()(auditConnector: DefaultAuditConnector) {
     sendExplicitAuditEvent(
       auditEventType = RulingCancelled,
       auditPayload = statusChangeAuditPayload(oldCase, updatedCase, operator)
+        + ("cancelReason" -> cancelReason(updatedCase))
```
gokyo: can be written in the line above (no need of a new line)

SteveSugden: I thought I've seen warnings about excessively long lines.  I prefer to keep it on the screen (unless we feel strongly about this)

PR #131, opened 2019-02-27
DIT-808 Include `justification` field in search

retroredge:


app/uk/gov/hmrc/tariffclassificationfrontend/views/advanced_search.scala.html
```diff
@@ -42,9 +42,9 @@ <h2 class="heading-medium">Enter search terms</h2>
   )
 
   @input_text(
-   field = form("good_description"),
-   label = "Goods description",
-   hint = Some("Searches the Ruling Goods Description and the Commercial Denomination"),
+   field = form("decision_details"),
+   label = "Ruling details",
+   hint = Some("Searches the Ruling Goods Description, justification and the Commercial Denomination"),
```
SteveSugden: Is that text per design?  Wondering why it's not "Justification"

ed0906:
Looks good to me, worth double checking with Steve and Aimee if you haven't already about the design as I seem to remember the hint being removed on the new design

app/uk/gov/hmrc/tariffclassificationfrontend/views/advanced_search.scala.html
retroredge: waiting for design

retroredge: I see what you're saying. I'll capitalise it for now but both the label and the hint will probably change when UX have a look at it

retroredge: Got the content now from UX. All lower cased :)

PR #132, opened 2019-02-27
DIT 249 - Take Ownership of a Case

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AppealCaseController.scala
```diff
@@ -42,7 +43,7 @@ class AppealCaseController @Inject()(override val authenticatedAction: Authentic
 
   override protected def redirect: String => Call = routes.CaseController.trader
 
-  override protected def isValidCase: Case => Boolean = { c: Case =>
+  override protected def isValidCase(c: Case)(implicit request: AuthenticatedRequest[_]): Boolean = {
```
ed0906: Had to change all of these, sorry in advance :(

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AssignCaseController.scala
```diff
@@ -0,0 +1,58 @@
...
+  override protected def redirect: String => Call = routes.CaseController.trader
+
+  override protected def isValidCase(c: Case)(implicit request: AuthenticatedRequest[_]): Boolean = c.queueId.isDefined && !c.assignee.map(_.id).contains(request.operator.id)
+
+  def get(reference: String): Action[AnyContent] = authenticatedAction.async { implicit request =>
+    getCaseAndRenderView(reference, c => successful(views.html.assign_case(c)))
+  }
+
+  def post(reference: String): Action[AnyContent] = authenticatedAction.async { implicit request =>
+    def respond: Case => Future[Result] = {
```
gokyo: How if I want to assign the case to myself after it was assigned to a different user?
Your code does not allow it now

ed0906: It’s not a requirement at the moment (I double checked with Joe)

gokyo: yes, but we will need it soon I believe ...

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
```diff
@@ -80,6 +80,14 @@ class CasesService @Inject()(appConfig: AppConfig,
     } yield updated
   }
 
+  def assignCase(original: Case, operator: Operator)
+                  (implicit hc: HeaderCarrier): Future[Case] = {
+    for {
+      updated <- connector.updateCase(original.copy(assignee = Some(operator)))
+      _ <- addAssignmentChangeEvent(original, updated, operator)
+      _ = auditService.auditCaseAssigned(updated, operator)
```
gokyo: do you mind changing the signature of `auditCaseAssigned` in `AuditService.scala`?
I think we should pass these arguments:
- Case c // the updated case
- Operator operator // the user that is logged in
- Operator assignee // the person that will work on that case

in this class `operator` will be the same as `assignee`, but at least we keep this definition more generic. In the future we will be able to have the assignee being different from the current logged in user.

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
```diff
@@ -236,6 +244,12 @@ class CasesService @Inject()(appConfig: AppConfig,
     addEvent(original, updated, details, operator)
   }
 
+  private def addAssignmentChangeEvent(original: Case, updated: Case, operator: Operator, comment: Option[String] = None)
+                                        (implicit hc: HeaderCarrier): Future[Unit] = {
```
gokyo: formatting

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
```diff
@@ -80,6 +80,14 @@ class CasesService @Inject()(appConfig: AppConfig,
     } yield updated
   }
 
+  def assignCase(original: Case, operator: Operator)
+                  (implicit hc: HeaderCarrier): Future[Case] = {
```
gokyo: formatting

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AssignCaseController.scala
```diff
@@ -0,0 +1,58 @@
...
+@Singleton
+class AssignCaseController @Inject()(authenticatedAction: AuthenticatedAction,
+                                     override val caseService: CasesService,
+                                     val messagesApi: MessagesApi,
+                                     override implicit val config: AppConfig) extends RenderCaseAction {
+
+
+  override protected def redirect: String => Call = routes.CaseController.trader
+
+  override protected def isValidCase(c: Case)(implicit request: AuthenticatedRequest[_]): Boolean = c.queueId.isDefined && !c.assignee.map(_.id).contains(request.operator.id)
```
gokyo: I know `Option.contains(x)` means  `== Some(x)`, but I find the `contains` confusing, especially if you are used to the `String.contains`.

How about rewriting the conditions with `==` or some other way?

gokyo: also, the method is now very long, let's define the body in a new line...

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AssignCaseController.scala
```diff
@@ -0,0 +1,58 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.concurrent.Future.successful
+
+@Singleton
+class AssignCaseController @Inject()(authenticatedAction: AuthenticatedAction,
+                                     override val caseService: CasesService,
+                                     val messagesApi: MessagesApi,
+                                     override implicit val config: AppConfig) extends RenderCaseAction {
+
```
gokyo: extra empty line

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AssignCaseController.scala
gokyo: I think you want to express the fact that the case is not assigned to anyone, or assigned to another user.
Right?

You could say:

```
override protected def isValidCase(c: Case)(implicit request: AuthenticatedRequest[_]): Boolean = {
    (c.queueId, c.assignee) match {
       case (Some(_), Some(id)) if id != request.operator.id => true
       case _ => false
    }
}
```

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
ed0906: What's wrong with it?

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
ed0906: What's wrong with it?

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AssignCaseController.scala
ed0906: That seems more readable, i'll change it

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
ed0906: I think the signature is fine as it is, as you can get the assignee from `c.assignee`
I'll change it so the `assigneeId` in the audit payload is the actual `assignee` though rather than the operator who actioned the change

ed0906: Pushed

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AssignCaseController.scala
ed0906: Pushed

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AssignCaseController.scala
ed0906: Fixed

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AssignCaseController.scala
ed0906: I suspect the UI will have a different pattern for that action - we can deal with it when we know

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
ed0906: This follows the pattern of every other method in that class

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
ed0906: This follows the pattern of every other method in that class

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
gokyo: the opening bracket of the implicit param is not parallel with the bracket above

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
gokyo: the opening bracket of the implicit param is not parallel with the bracket above

PR #133, opened 2019-02-28
Bau remove uktt check for mandatory validation

retroredge:


app/uk/gov/hmrc/tariffclassificationfrontend/forms/DecisionFormData.scala
```diff
@@ -46,7 +46,7 @@ class DecisionForm @Inject()(commodityCodeConstraints: CommodityCodeConstraints)
 
   val mandatoryFieldsForm: Form[DecisionFormData] = Form(
     mapping(
-      "bindingCommodityCode" -> text.verifying(commodityCodeConstraints.commodityCodeExistsInUKTradeTariff),
```
ed0906: Does this mean we don't need to inject the form into the controller any more?

PR #140, opened 2019-03-07
DIT-758: introducing authentication

ed0906:


test/unit/uk/gov/hmrc/tariffclassificationfrontend/connector/RulingConnectorTest.scala
```diff
@@ -45,11 +45,9 @@ class RulingConnectorTest extends UnitSpec with WithFakeApplication with Wiremoc
 
     "POST to the Ruling Store" in {
       given(config.rulingUrl).willReturn(wireMockUrl)
-      given(config.apiToken).willReturn("auth")
 
       stubFor(
         post("/binding-tariff-rulings/ruling/id")
-          .withHeader("X-Api-Token", equalTo("auth"))
```
ed0906: Removed this as it's part of the verify above

app/uk/gov/hmrc/tariffclassificationfrontend/connector/AuthenticatedHttpClient.scala
```diff
@@ -0,0 +1,88 @@
...
+import play.api.libs.ws.WSClient
+import uk.gov.hmrc.http.{HeaderCarrier, HttpResponse}
+import uk.gov.hmrc.play.audit.http.connector.AuditConnector
+import uk.gov.hmrc.play.bootstrap.http.DefaultHttpClient
+import uk.gov.hmrc.tariffclassificationfrontend.config.AppConfig
+
+import scala.concurrent.Future
+
+class AuthenticatedHttpClient @Inject()(auditConnector: AuditConnector, wsClient: WSClient, actorSystem: ActorSystem)
+                                       (implicit val config: AppConfig)
```
gokyo: do you prefer `config` explicit or implicit?

app/uk/gov/hmrc/tariffclassificationfrontend/connector/AuthenticatedHttpClient.scala
```diff
@@ -0,0 +1,88 @@
...
+
+}
+
+trait InjectAuthHeader {
+
+  private val headerName: String = "X-Api-Token"
+
+  def addAuth(implicit config: AppConfig, hc: HeaderCarrier): HeaderCarrier = {
+    hc.headers.toMap.get(headerName) match {
+      case Some(_) => hc
```
gokyo: this case makes sure the header is injected only if not present (if two of our back-end microservices talk to each other, the header should be propagated already - when we use http-verbs)

app/uk/gov/hmrc/tariffclassificationfrontend/connector/FileStoreConnector.scala
```diff
@@ -64,6 +63,7 @@ class FileStoreConnector @Inject()(appConfig: AppConfig, http: HttpClient, ws: W
     )
 
     ws.url(s"${appConfig.fileStoreUrl}/file")
+      .withHeaders(http.addAuth(appConfig, hc).headers: _*)
```
gokyo: In this case we cannot use `http-verbs`, so this needs to be explicitly passed like this
We need to do the same in the other microservices (`binding-tariff-admin-frontend` + `binding-tariff-trader-frontend`) uploading files to `binding-tariff-filestore`

app/uk/gov/hmrc/tariffclassificationfrontend/connector/RulingConnector.scala
```diff
@@ -17,20 +17,17 @@
...
 import scala.concurrent.Future
 
 @Singleton
-class RulingConnector @Inject()(configuration: AppConfig,
-                                http: HttpClient) {
+class RulingConnector @Inject()(configuration: AppConfig, http: AuthenticatedHttpClient) {
 
   def notify(id: String)(implicit hc: HeaderCarrier): Future[Unit] = {
-    val authenticatedHeaders: HeaderCarrier = hc.withExtraHeaders("X-Api-Token" -> configuration.apiToken)
-    http.POSTEmpty(s"${configuration.rulingUrl}/binding-tariff-rulings/ruling/$id")(HttpReads.readRaw, authenticatedHeaders, global).map(_ => ())
```
gokyo: the implicits can be removed now

test/unit/uk/gov/hmrc/tariffclassificationfrontend/connector/FileStoreConnectorSpec.scala
```diff
@@ -190,9 +186,12 @@ class FileStoreConnectorSpec extends UnitSpec with WiremockTestServer with Mocki
     val file = FileUpload(TemporaryFile("example-file.txt"), "file.txt", "text/plain")
     val result = await(connector.upload(file))
 
-    verify(postRequestedFor(urlEqualTo("/file"))
-      .withRequestBody(containing("file"))
-      .withRequestBody(containing("publish"))
+    verify(
+      postRequestedFor(
+        urlEqualTo("/file"))
+        .withHeader("X-Api-Token", equalTo(fakeAuthToken))
```
gokyo: In this case we do not use `http-verbs` - I had to use a different instance of `AppConfig` 
Let me know if you have a better idea here

test/unit/uk/gov/hmrc/tariffclassificationfrontend/connector/EmailConnectorSpec.scala
```diff
@@ -75,11 +56,16 @@ class EmailConnectorSpec extends UnitSpec
...
-            .withBody(fromResource("email_template-response.json"))
+          .withBody(fromResource("email_template-response.json"))
           .withStatus(HttpStatus.SC_OK))
       )
 
       await(connector.generate(email)) shouldBe EmailTemplate("text", "html", "from", "subject", "service")
+
+      verify(
+        postRequestedFor(urlEqualTo(s"/templates/${EmailType.COMPLETE}"))
+          .withoutHeader("X-Api-Token")
```
gokyo: calls to microservices not belonging to us should NOT contain the extra header we inject

test/unit/uk/gov/hmrc/tariffclassificationfrontend/connector/EmailConnectorSpec.scala
```diff
@@ -16,56 +16,37 @@
...
         .willReturn(aResponse()
           .withStatus(HttpStatus.SC_ACCEPTED))
       )
 
-      await(connector.send(email))
+      await(connector.send(email)) shouldBe ((): Unit)
+
+      verify(
+        postRequestedFor(urlEqualTo("/hmrc/email"))
+          .withoutHeader("X-Api-Token")
```
gokyo: `without`

test/unit/uk/gov/hmrc/tariffclassificationfrontend/connector/ConnectorTest.scala
```diff
@@ -0,0 +1,66 @@
...
+import uk.gov.tariffclassificationfrontend.utils.{ResourceFiles, WiremockTestServer}
+
+abstract class ConnectorTest extends UnitSpec with WithFakeApplication
+  with WiremockTestServer with MockitoSugar with ResourceFiles
+  with BeforeAndAfterAll {
+
+  private val actorSystem = ActorSystem.create("testActorSystem")
+
+  protected implicit val realConfig: AppConfig = fakeApplication.injector.instanceOf[AppConfig]
+  protected val appConfig: AppConfig = mock[AppConfig]
```
gokyo: this is used when asserting the auth header when we upload files (where we do not use http-verbs)

app/uk/gov/hmrc/tariffclassificationfrontend/connector/AuthenticatedHttpClient.scala
ed0906: Doesnt matter, usually with constructors we have it explicit

app/uk/gov/hmrc/tariffclassificationfrontend/connector/FileStoreConnector.scala
ed0906: Could we not just add something to `AuthenticatedHttpClient` e.g. 
`lazy val apiKeyHeader ("X-Api-Key" -> "xyz")`

Which we can call in instances like this using `http.apiKeyHeader`

test/unit/uk/gov/hmrc/tariffclassificationfrontend/connector/FileStoreConnectorSpec.scala
ed0906: Thats fine, this connector method is an exception to the others as it uses WS

app/uk/gov/hmrc/tariffclassificationfrontend/connector/FileStoreConnector.scala
gokyo: Remember that for this specific request we are not using `http-verbs`, but WsClient instead.
We could use the `wsClient` field inside  `AuthenticatedHttpClient`. I can give it a go

gokyo: do you prefer this ?
```
    ws.url(s"${appConfig.fileStoreUrl}/file")
      .withHeaders( hc.headers: _* )
      .withHeaders( http.authHeaders(appConfig) )
      .post(Source(List(filePart, dataPart)))
      .map(response => Json.fromJson[FileMetadata](Json.parse(response.body)).get)
  }
```

ed0906: Does the job :)

PR #142, opened 2019-03-13
DIT-815 update commodity code error message

retroredge:


ed0906:
Presume acceptance tests dont need changing?

PR #144, opened 2019-03-15
Refactor queue name map

SteveSugden:
Small attempts at increasing code coverage


SteveSugden:
Private PR for @pllaurado

PR #163, opened 2019-03-27
DIT-899: do not show cancel-ruling button for expired rulings

gokyo:


app/uk/gov/hmrc/tariffclassificationfrontend/models/Case.scala
```diff
@@ -34,4 +34,7 @@ case class Case
   decision: Option[Decision],
   attachments: Seq[Attachment],
   keywords: Set[String] = Set.empty
-)
+) {
+
+  def rulingHasNotExpired: Boolean = decision.flatMap(_.effectiveEndDate).exists(_.compareTo(Instant.now) >= 0)
```
ed0906: Needs a test in `CaseTest`

PR #164, opened 2019-03-27
DIT-918 Move the Boards File Number to the trader tab

retroredge:


app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/case_trader.scala.html
```diff
@@ -84,7 +84,18 @@
...
             </div>
 
-            @if(btiApp.agent.isDefined) {
+        @if(c.caseBoardsFileNumber.isDefined) {
+            <div class="grid-row">
+                <div class="column-one-third">
+                    <p id="boards-file-number-label">@messages("case.application.boards-file-number")</p>
+                </div>
+                <div class="column-two-thirds">
+                    <p id="boards-file-number">@{c.caseBoardsFileNumber}</p>
```
gokyo: I believe you need to extract the value from the option (`.get`)

retroredge: It seems to work without a call to `.get` in a Twirl but I've added one anyway

PR #168, opened 2019-03-29
DIT-904 - thumbnail open new tab

SteveSugden:
Created new 'thumbnail' partial view which includes anchor tag opening image in new window/tab

This view can potentially be re-used to display attachment thumbnails on the 'attachments' tab


app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/advanced_search_results.scala.html
```diff
@@ -90,11 +91,12 @@
      <span id="advanced_search_results-row-@{index}-keywords">@result.c.keywords.mkString(", ")</span>
     </div>
 
+    @*Attached Images*@
```
pllaurado: I would prefer to have a component called display_attached_images or something  similar  rather than a comment :)

SteveSugden: ?  I added the comment to match the others for the different items in the search results

pllaurado: It was a suggestion/preference, easier to test and maintain but feel free to resolve conversation.

PR #169, opened 2019-03-29
DIT-823 Show Case Statuses in a Lozenge

ed0906:


gokyo:
Do we need to change the layout of the case status?
I think we need to speak with @mcgrathjoe

ed0906:
> Do we need to change the layout of the case status?
> I think we need to speak with @mcgrathjoe

Change the layout in what way?

SteveSugden:
Demo'd changes to Joe and Vickie.  Discussed the possibility of colour-coding the case statuses but agreed that this would be a future ticket.  (This will be easier now as the colour logic will just need to be implemented in the case_status html component).

PR #172, opened 2019-04-04
DIT-903 Add pre-wrap style to fields that are potential multiline

retroredge:


retroredge:
> Code changes look good to me. Should we demo before merging or after?

Before I think

PR #183, opened 2019-04-17
DIT 961 - SLA Report

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/models/CaseReportFilter.scala
```diff
@@ -0,0 +1,52 @@
...
+object CaseReportFilter {
+
+  val decisionStartKey = "decision_start"
+
+  implicit def binder(implicit rangeBinder: QueryStringBindable[InstantRange]): QueryStringBindable[CaseReportFilter] = new QueryStringBindable[CaseReportFilter] {
+
+    override def bind(key: String, requestParams: Map[String, Seq[String]]): Option[Either[String, CaseReportFilter]] = {
+      implicit val rp: Map[String, Seq[String]] = requestParams
+
+      val decisionStart: Option[InstantRange] = rangeBinder.bind(decisionStartKey, requestParams).filter(_.isRight).map(_.right.get)
```
SteveSugden: Same comment as BE PR.  Is there a better way of combining the filter and the map?

app/uk/gov/hmrc/tariffclassificationfrontend/views/forms/error_summary.scala.html
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@(form: Form[_])(implicit messages: Messages)
+
+@if(form.hasErrors) {
+ <div id="sla_criteria-errors" class="error-summary" role="alert" aria-labelledby="sla_criteria-error" tabindex="-1">
```
SteveSugden: Is this component specific to errors on SLA reports?   If so, should the file name reflect this?

app/uk/gov/hmrc/tariffclassificationfrontend/views/report_sla.scala.html
```diff
@@ -0,0 +1,137 @@
...
+            <dt class="column-one-quarter">
+                To
+            </dt>
+            <dd id="report_sla-to_date" class="column-three-quarters">
+                @Dates.format(filter.max)
+            </dd>
+        </div>
+    </dl>
+
+    @defining(Seq(
```
SteveSugden: I was expecting to see the processing of the report data (histograms) pulled out into some sort of SLAReport object with the UI just responsible for rendering it.  Won't we have to repeat this logic for the 'csv' version of the report?

app/uk/gov/hmrc/tariffclassificationfrontend/models/CaseReportFilter.scala
ed0906: I dont think there is i'm afraid :(

app/uk/gov/hmrc/tariffclassificationfrontend/views/forms/error_summary.scala.html
ed0906: Nah it's copied from the trader UI

app/uk/gov/hmrc/tariffclassificationfrontend/views/report_sla.scala.html
ed0906: Yep, we can pull any duplicates out as part of that. I suspect we can just wrap the histogram in a class with some useful methods specific to the report

ed0906: @SteveSugden I've pulled this out into an SLAReport class. Should make the CSV bit a lot easier

PR #188, opened 2019-04-25
DIT-975 Add `explanation` field to ruling

retroredge:


app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/ruling/ruling_explanation.scala.html
```diff
@@ -0,0 +1,55 @@
...
+
+@if(c.decision.isDefined) {
+    @defining(c.decision.get) { d: Decision =>
+        <h3 class="heading-medium">@messages("case.decision.holder.information")</h3>
+        <div class="grid-row" id="information-for-holder">
+            <dt class="column-one-third">
+                @messages("case.decision.explanation")
+            </dt>
+            <dd class="column-two-thirds">
+            @if(d.explanation.exists(_.trim.nonEmpty)) {
```
ed0906: Do we need the `_trim.nonEmpty` if we mandate that is always defined+nonempty with form validation?

aaroncarroll82: I think it needs to be there when looking at a case that may have existed before the change as it would have effectively bypassed that validation

ed0906: If the case existed before this change it will have explanation set to `None` :)

PR #189, opened 2019-04-25
DIT 963 - Referral Report

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/models/ReportResult.scala
```diff
@@ -20,4 +20,7 @@ case class ReportResult
 (
   group: Option[String],
   value: Seq[Int]
-)
+) {
+  def size: Int = value.size
+  def average: Double = value.sum.toDouble / size
```
SteveSugden: Do we need to guard against size being zero?

app/uk/gov/hmrc/tariffclassificationfrontend/views/ReferralReport.scala
```diff
@@ -0,0 +1,30 @@
...
+
+package uk.gov.hmrc.tariffclassificationfrontend.views
+
+import uk.gov.hmrc.tariffclassificationfrontend.models.{Queue, ReportResult}
+
+class ReferralReport(results: Seq[ReportResult]) {
+
+  lazy val count: Int = results.map(_.size).sum
+
+  lazy val average: Int = Math.round(results.flatMap(_.value).sum.toDouble / count).toInt
```
SteveSugden: Again, can count be zero?

app/uk/gov/hmrc/tariffclassificationfrontend/models/ReportResult.scala
ed0906: Count will never be zero :) there wouldnt be a ReportResult for it if it was

app/uk/gov/hmrc/tariffclassificationfrontend/views/ReferralReport.scala
ed0906: Count will never be zero :) there wouldnt be a ReportResult for it if it was

SteveSugden: I thought that'd be the case, but worth asking the question.

PR #192, opened 2019-04-30
DIT-947 Case Status Lozenge Colouring

ed0906:


test/unit/uk/gov/hmrc/tariffclassificationfrontend/views/partials/statuses/CaseStatusViewSpec.scala
```diff
@@ -27,11 +27,24 @@ class CaseStatusViewSpec extends ViewSpec {
 
   "Case Status" should {
 
+    "render CANCELLED without the reason code" in {
```
SteveSugden: Is this just to test the logic?  Cancelled should always have a reason.

ed0906: Yeh but from a data model perspective it isn't essential, and we still want it to show in Grey

ed0906: It just saves Testers any headaches from changing the status in the DB and wondering why it still shows in blue

ed0906:
> Looks good. SomeCaseStatusViewSpec tests that the blue/gray styles are applied for the right status?

I've added some new assertions for the class :)

PR #193, opened 2019-04-30
Dit 832 ruling pdf

SteveSugden:
PR for half of DIT 832 (ruling PDF)


app/uk/gov/hmrc/tariffclassificationfrontend/views/templates/ruling_template.scala.html
```diff
@@ -18,7 +18,7 @@
 @import uk.gov.hmrc.tariffclassificationfrontend.models.{Case, Decision}
 @import uk.gov.hmrc.tariffclassificationfrontend.utils.Dates
 
-@(c: Case)(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
+@(c: Case, d: Decision)(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
```
SteveSugden: Changed to pass decision into the template so that I could just copy/paste the template body from trader version (to ensure they are kept in line)

PR #194, opened 2019-05-01
Dit 832 - application pdf resync

pllaurado:


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/PdfDownloadController.scala
```diff
@@ -49,10 +49,13 @@ class PdfDownloadController @Inject()(authenticatedAction: AuthenticatedAction,
 
   def applicationPdf(reference: String): Action[AnyContent] = authenticatedAction.async { implicit request =>
     caseService.getOne(reference) flatMap {
-      case Some(c: Case) =>
-        fileStore.getAttachments(c).flatMap { files =>
-          generatePdf(application_template(c, files), s"BTIApplication_$reference.pdf")
-        }
+      case Some(c: Case) => {
+        fileStore.getAttachments(c).flatMap(files =>
+          fileStore.getLetterOfAuthority(c).flatMap(letter =>
```
SteveSugden: Should this not be done in parallel (like the trader Ui)?

app/uk/gov/hmrc/tariffclassificationfrontend/views/templates/application_template.scala.html
```diff
@@ -73,6 +73,14 @@ <h2 id=@headerId>@messages(headerId)</h2>
                 .pre-wrap {
```
SteveSugden: Is the html part of this file a copy/paste from the traderUI?   (It seems to be - same duplicate ids - which is good)

conf/messages
```diff
@@ -186,43 +186,47 @@ pdf.ruling.section.ruling.legalInformation.section2.paragraph2.listitem1 = infor
 pdf.ruling.created.label = File created
 
 
-pdf.application.title = Application for a goods classification ruling
-pdf.application.submitted = Application submitted on
-pdf.application.reference = Application reference number
-pdf.application.section.applicant.heading = Applicant details
+pdf.application.title = Your application for a Binding Tariff Information ruling
```
SteveSugden: Similar - I assume you copied all of the "pdf.application...." messages from one project to the other

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/PdfDownloadController.scala
pllaurado: I've changed into this approach which allows me easier to redirect NOT FOUND. Ultimately the for comprension is translating into the same structure. Is just nicer to read.

app/uk/gov/hmrc/tariffclassificationfrontend/views/templates/application_template.scala.html
pllaurado: yes it is

app/uk/gov/hmrc/tariffclassificationfrontend/views/templates/application_template.scala.html
```diff
@@ -116,154 +130,189 @@ <h2 id=@headerId>@messages(headerId)</h2>
...
+                            @agent.eoriDetails.addressLine3 <br/>
+                            @agent.eoriDetails.postcode <br/>
+                            @agent.eoriDetails.country
+                        </td>
+                    </tr>
+                    <tr>
+                        <td> Written authorisation </td>
+                        <td id="agentDetails.letterOfAuthority">
+                        @if(letterOfAuthority.isDefined){@letterOfAuthority.get.fileName}
+                </td>
```
retroredge: Formatting is a bit odd here

conf/messages
pllaurado: Why not?? I think we should move it anyway... Not big fan to put into the Back End, but at least would be just one implementation.

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/PdfDownloadController.scala
SteveSugden: You are right.  I thought the for comprehension structure automatically resulted in parallel execution.  My bad.     Would it not be better to re-structure like this https://alvinalexander.com/scala/how-use-multiple-scala-futures-in-for-comprehension-loop . (in both UIs).

Not critical I guess as generating the pdf takes far longer than getting the data together to make the request to the pdf-service and printing pdfs is not something that will be done very often.

ed0906: I'd use for comprehension here rather than `flatMap`

app/uk/gov/hmrc/tariffclassificationfrontend/views/templates/application_template.scala.html
```diff
@@ -116,154 +130,189 @@ <h2 id=@headerId>@messages(headerId)</h2>
...
+                                @row.fileName <br/>
+                            }
+                        }
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>@messages("pdf.application.section.aboutItem.importOrExport")</td>
+                        <td id="application.importOrExport">
+                        @{application.importOrExport match {
+                            case Some("IMPORT") => messages("pdf.application.section.aboutItem.import")
```
retroredge: Would this ever need to be a case insensitive match?

conf/messages
SteveSugden: I wasn't suggesting that was wrong.   Just checking that the content (text) was the same in both places

app/uk/gov/hmrc/tariffclassificationfrontend/views/templates/application_template.scala.html
ed0906: Same can be achieved with `letterOfAuthority.map(_.fileName)`

app/uk/gov/hmrc/tariffclassificationfrontend/views/templates/application_template.scala.html
ed0906: It's an Enum I think

ed0906: Should the match not be on the enum value anyway?

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/PdfDownloadController.scala
pllaurado: I use for comprehension but I've change it in order to align both implementations just because **one of them use case Not found** and was easier in this way.

ed0906: `Easier != better` though @pllaurado

app/uk/gov/hmrc/tariffclassificationfrontend/views/templates/application_template.scala.html
retroredge: I say 'yes' to Ed's last comment

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/PdfDownloadController.scala
pllaurado: Easier (not always means) better would be more accurate.

app/uk/gov/hmrc/tariffclassificationfrontend/views/templates/application_template.scala.html
pllaurado: done

PR #196, opened 2019-05-03
DIT 392 - Show Commodity Code Expiry Dates

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/service/CommodityCodeService.scala
```diff
@@ -16,26 +16,56 @@
...
+
+  private def padTo10Digits(input: String): String = {
+    val trimmed = input.trim
+    if (trimmed.length > padLimit) trimmed.substring(0, padLimit)
+    else trimmed.padTo[Char, String](padLimit, '0').mkString
+  }
+
+  private val commodityCodesFromFile: Seq[CommodityCode] = {
+    val url = getClass.getClassLoader.getResource(appConfig.commodityCodePath)
+    val lines = Source.fromURL(url, "UTF-8").getLines()
```
aaroncarroll82: Should we be closing this file after reading? This will leave a handle to file open till the JVM exists.

app/uk/gov/hmrc/tariffclassificationfrontend/service/CommodityCodeService.scala
```diff
@@ -16,26 +16,56 @@
...
+          CommodityCode(commodityCode)
+        }
+      }.toSeq
   }
 
-  private lazy val commodityCodesFromFile: Seq[String] = {
-    val url = getClass.getClassLoader.getResource("commodityCodes.txt")
-    (for (line <- Source.fromURL(url, "UTF-8").getLines()) yield line).toSeq
+  private def split(string: String): Array[String] = {
+    string.split(";").map(_.replaceAll("\"", ""))
```
aaroncarroll82: Is it worth having the separator as something configurable? If a new csv is generated in the future it might actually use commas potentially (as this current one is actually a semi-colon separated file rather than a comma separated file)

app/uk/gov/hmrc/tariffclassificationfrontend/models/CommodityCode.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import java.time.{Clock, Instant}
+
+case class CommodityCode
+(
+  code: String,
+  expiry: Option[Instant] = None
+) {
+  def isExpired(implicit clock: Clock): Boolean = expiry.exists(_.isBefore(clock.instant()))
+  def isLive(implicit clock: Clock): Boolean = !isExpired
```
pllaurado: just something curious:  if expiry does not exist the assumption is that the code is forever alive?

app/uk/gov/hmrc/tariffclassificationfrontend/forms/CommodityCodeConstraints.scala
```diff
@@ -16,16 +16,24 @@
...
+import java.time.Clock
+
 import javax.inject.{Inject, Singleton}
 import play.api.data.validation.{Constraint, Invalid, Valid}
+import uk.gov.hmrc.tariffclassificationfrontend.config.AppConfig
 import uk.gov.hmrc.tariffclassificationfrontend.service.CommodityCodeService
 
 @Singleton
-class CommodityCodeConstraints @Inject()(commodityCodeService: CommodityCodeService) {
+class CommodityCodeConstraints @Inject()(commodityCodeService: CommodityCodeService, appConfig: AppConfig) {
```
pllaurado: Good idea 👍

app/uk/gov/hmrc/tariffclassificationfrontend/service/CommodityCodeService.scala
```diff
@@ -16,26 +16,56 @@
...
+    commodityCodesFromFile.find(_.code == canonicalCode)
+  }
+
+  private def padTo10Digits(input: String): String = {
+    val trimmed = input.trim
+    if (trimmed.length > padLimit) trimmed.substring(0, padLimit)
+    else trimmed.padTo[Char, String](padLimit, '0').mkString
+  }
+
+  private val commodityCodesFromFile: Seq[CommodityCode] = {
```
pllaurado: Why is not lazy anymore? Do we want to read the file every time?

app/uk/gov/hmrc/tariffclassificationfrontend/service/CommodityCodeService.scala
pllaurado: From : https://alvinalexander.com/scala/how-to-open-read-text-files-in-scala-cookbook-examples

`Read the file into a string
Use this approach to get all of the lines from the file as one String:

val fileContents = Source.fromFile(filename).getLines.mkString
This approach has the side effect of leaving the file open as long as the JVM is running, but for short-lived shell scripts, this shouldn’t be an issue; the file is closed when the JVM shuts down.

Properly closing the file
To properly close the file, get a reference to the BufferedSource when opening the file, and manually close it when you’re finished with the file:

val bufferedSource = Source.fromFile("example.txt")
for (line <- bufferedSource.getLines) {
    println(line.toUpperCase)
}

bufferedSource.close
For automated methods of closing the file, see the “Loan Pattern” examples in the Discussion.

The getLines method of the Source class returns a scala.collection.Iterator. The iterator returns each line without any newline characters. An iterator has many methods for working with a collection, and for the purposes of working with a file, it works well with the for loop, as shown.

 `

ed0906: I suspect the stream is closed when the iterator finishes. If I try and close it manually I get IOExceptions as the stream is closed

app/uk/gov/hmrc/tariffclassificationfrontend/models/CommodityCode.scala
ed0906: Correct :)

app/uk/gov/hmrc/tariffclassificationfrontend/service/CommodityCodeService.scala
ed0906: Hmmm, good suggestion, it's not really a requirement at the moment though. I think the long term solution is to get the codes from an API

app/uk/gov/hmrc/tariffclassificationfrontend/service/CommodityCodeService.scala
ed0906: It's a val so it would only be read once.
I've added lazy back on anyway

ed0906:
> +24,929 −15,939 is this a normal day at work? xD

If only I could claim all those

app/uk/gov/hmrc/tariffclassificationfrontend/service/CommodityCodeService.scala
aaroncarroll82: This works (not saying it's the best ever code) - 

val src = Source.fromURL(url, "UTF-8")
    val lines = src.getLines()

    val byHeader: Map[String, Int] = split(lines.next()).zipWithIndex.toMap

    try {
      lines
        .map(split)
        .filter(columns => columns(byHeader("leaf")) == "1")
        .map { columns =>
          val commodityCode = columns(byHeader("goods_nomenclature_item_id"))
          val expiry = columns(byHeader("validity_end_date"))
          if (expiry.nonEmpty) {
            val date = LocalDateTime.parse(expiry, dateTimeFormatter).atZone(appConfig.clock.getZone).toInstant
            CommodityCode(commodityCode, Some(date))
          } else {
            CommodityCode(commodityCode)
          }
        }.toSeq
    } finally {
      println("OK I will close the file then")
      src.close()
    }

Without that I can see using `sudo lsof -u aaroncarroll | grep 'commodityCodes'` that it's left open otherwise

app/uk/gov/hmrc/tariffclassificationfrontend/service/CommodityCodeService.scala
```diff
@@ -16,26 +16,56 @@
...
+        } else {
+          CommodityCode(commodityCode)
+        }
+      }.toSeq
   }
 
-  private lazy val commodityCodesFromFile: Seq[String] = {
-    val url = getClass.getClassLoader.getResource("commodityCodes.txt")
-    (for (line <- Source.fromURL(url, "UTF-8").getLines()) yield line).toSeq
+  private def split(string: String): Array[String] = {
```
retroredge: `splitAndRemoveQuotes` or some name to convey that the function is doing more than just splitting the string

app/uk/gov/hmrc/tariffclassificationfrontend/service/CommodityCodeService.scala
```diff
@@ -16,26 +16,56 @@
...
+    val trimmed = input.trim
+    if (trimmed.length > padLimit) trimmed.substring(0, padLimit)
+    else trimmed.padTo[Char, String](padLimit, '0').mkString
+  }
+
+  private lazy val commodityCodesFromFile: Seq[CommodityCode] = {
+    val url = getClass.getClassLoader.getResource(appConfig.commodityCodePath)
+    val lines = Source.fromURL(url, "UTF-8").getLines()
+
+    val byHeader: Map[String, Int] = split(lines.next()).zipWithIndex.toMap
```
retroredge: `split` is called here and on line 53, are both necessary?

app/uk/gov/hmrc/tariffclassificationfrontend/service/CommodityCodeService.scala
```diff
@@ -16,26 +16,56 @@
...
-    val canonicalCode: String =
-      if (commodityCode.length > padLimit) commodityCode.substring(0, padLimit)
-      else commodityCode.trim.padTo[Char, String](padLimit, '0').mkString
+  def find(commodityCode: String): Option[CommodityCode] = {
+    val canonicalCode: String = padTo10Digits(commodityCode)
+
+    commodityCodesFromFile.find(_.code == canonicalCode)
+  }
+
+  private def padTo10Digits(input: String): String = {
```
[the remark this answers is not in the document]
ed0906: This is for the form input (as before) not the CSV. Nothings changed I just pulled it into a method

PR #197, opened 2019-05-08
DIT-774 Split my cases between open and referred statuses

pllaurado:


app/uk/gov/hmrc/tariffclassificationfrontend/views/my_cases.scala.html
```diff
@@ -16,11 +16,21 @@
...
+@import uk.gov.hmrc.tariffclassificationfrontend.models.{Case, CaseStatus, Operator, Paged, Queue}
 @import uk.gov.hmrc.tariffclassificationfrontend.views.html.includes.main
 @import uk.gov.hmrc.tariffclassificationfrontend.views.html.partials.{cases_list, nav_search, queue_nav}
 @(queues: Seq[Queue], cases: Paged[Case], operator: Operator)(implicit request: AuthenticatedRequest[_], messages: Messages, appConfig: AppConfig)
 
+@referredStatuses = @{
+    Seq(CaseStatus.REFERRED, CaseStatus.SUSPENDED)
+}
+@nonReferredCases = @{
+    cases.copy(results = cases.results.filter(c => !referredStatuses.contains(c.status)))
```
ed0906: I know we dont have pagination on the cases list, but if we add it later it will be messed up by this filter.

I would change the signature of cases_list to accept a `Seq[Case]` instead

PR #201, opened 2019-05-13
DIT-775 adding totals in each queue

pllaurado:


app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
```diff
@@ -230,6 +231,15 @@ class CasesService @Inject()(appConfig: AppConfig,
...
   }
 
+  def countCasesByQueue(operator: Operator)(implicit hc: HeaderCarrier): Future[Map[String, Int]] = {
+    for {
+      countMyCases <- getCasesByAssignee(operator, NoPagination())
+      countByQueue <- reportingService.getQueueReport
+    } yield countByQueue.map(r => (
+          r.group.getOrElse(Queues.gateway.id), r.value.size)).toMap ++
+          Map("my-cases" -> countMyCases.size)
+  }
```
aaroncarroll82: Would it be a bit more readable if the mapping to final result was completed inside the for comprehension? Something like this:
def countCasesByQueue(operator: Operator)(implicit hc: HeaderCarrier): Future[Map[String, Int]] = {
    for {
      countMyCases <- getCasesByAssignee(operator, NoPagination())
      countByQueue <- reportingService.getQueueReport
      queueIdAndCount = countByQueue.map(r => (
        r.group.getOrElse(Queues.gateway.id), r.value.size)).toMap + ("my-cases" -> countMyCases.size)
    } yield queueIdAndCount
  }

app/uk/gov/hmrc/tariffclassificationfrontend/service/ReportingService.scala
```diff
@@ -39,8 +40,19 @@ class ReportingService @Inject()(connector: BindingTariffClassificationConnector
     connector.generateReport(report)
   }
 
+  def getQueueReport(implicit hc: HeaderCarrier): Future[Seq[ReportResult]] = {
+    val statuses = Set(NEW, OPEN, REFERRED, SUSPENDED).map(_.toString)
+
+    val report = CaseReport(
+      filter = CaseReportFilter(status = Some(statuses), assigneeId = Some("none")),
```
aaroncarroll82: This looks a bit odd - can't we use assigneeId = None, rather than Some("none")

pllaurado: maybe looks odd because you are mixing None (optional scala object) and none (which means, select cases with no assigned operator) :-)

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
pllaurado: This one already looks good to me, actually the name of the actual method (**countCasesByQueue**) should be like the new var you have created (**queueIdAndCount**). Maybe is that scala madness...

PR #202, opened 2019-05-13
Dit 906

retroredge:


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AttachmentsController.scala
```diff
@@ -44,14 +44,47 @@ class AttachmentsController @Inject()(verify: RequestActions,
                                       fileService: FileStoreService,
                                       val messagesApi: MessagesApi,
                                       implicit val appConfig: AppConfig,
-                                      implicit val mat: Materializer) extends FrontendController with I18nSupport {
+                                      implicit val mat: Materializer) extends RenderCaseAction with I18nSupport {
```
aaroncarroll82: Do you need to have the with I18nSupport here now? It's pulled in by theRenderCaseAction

retroredge: Not added that bit :)

retroredge: All of the controllers use that trait. I guess if we ever provide other languages we'd need all this

app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/attachments_list.scala.html
```diff
@@ -52,9 +59,18 @@
...
                                 </td>
                             }
+
                             <td id="@{id}-row-@{index}-date" class="w-15">
                                 @Dates.format(file.timestamp)
                             </td>
+
+                            @if(showRemoval &&
+                                caseRef.isDefined &&
+                                request.hasPermission(Permission.REMOVE_ATTACHMENTS)) {
```
aaroncarroll82: Should this link always be here for every attachment? The ticket mentions about an attachment that's already published should replace the remove link with some text that says 'Published'

conf/messages
```diff
@@ -158,6 +158,7 @@ reassign-case.submit.button = Move back to queue
 suppress_case.form.state.required = Select yes if you have contacted the applicant
 reject_case.form.state.required = Select yes if you have contacted the applicant
 refer_case.form.state.required = Select yes if you have contacted the applicant
+remove_attachment.form.state.required = Select yes if you want to remove the attachment
```
aaroncarroll82: Minor possible usability/consistency issue. The error message suggests choose 'yes' to confirm, but the option is actually labelled 'Yes' (with capital Y). Quite pedantic, but should it align?
Screenshot 2019-05-14 at 16 20 17

app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/attachments_list.scala.html
aaroncarroll82: After a talk through about this - it's not part of the AC at the moment for this ticket, so not an issue. In the future it could possibly be derived from a combo of existing public flag on the attachment and case status of completed (to hide remove link)

conf/messages
aaroncarroll82: Noticed this is a common pattern throughout the application so probably not a problem.

retroredge: we don't normally capitalise words mid-sentence

retroredge: checked with Aimee and that is the required text

PR #204, opened 2019-05-15
Show filtered activities on activity tab, and sample tab

aaroncarroll82:
Ready to go

### DIFF
--- a/app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
+++ b/app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
@@ -24,6 +24,7 @@ import play.api.mvc.QueryStringBindable
 import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.tariffclassificationfrontend.config.AppConfig
 import uk.gov.hmrc.tariffclassificationfrontend.models.CaseStatus._
+import uk.gov.hmrc.tariffclassificationfrontend.models.EventType.EventType
 import uk.gov.hmrc.tariffclassificationfrontend.models._
 import uk.gov.hmrc.tariffclassificationfrontend.models.request.NewEventRequest
 import uk.gov.hmrc.tariffclassificationfrontend.utils.JsonFormatters._
@@ -75,7 +76,17 @@ class BindingTariffClassificationConnector @Inject()(appConfig: AppConfig, clien
   }
 
   def findEvents(reference: String, pagination: Pagination)(implicit hc: HeaderCarrier): Future[Paged[Event]] = {
-    val url = s"${appConfig.bindingTariffClassificationUrl}/cases/$reference/events?page=${pagination.page}&page_size=${pagination.pageSize}"
+    findFilteredEvents(reference,pagination, None)
+  }
+
+  def findFilteredEvents(reference: String, pagination: Pagination, onlyEventTypes: Option[Set[EventType]])(implicit hc: HeaderCarrier): Future[Paged[Event]] = {
+
+    val searchParam = onlyEventTypes match {
+      case Some(only) => s"case_reference=$reference${only.mkString("&type=","&type=","&")}"
+      case None => s"case_reference=$reference&"
+    }
+
+    val url = s"${appConfig.bindingTariffClassificationUrl}/events?${searchParam}page=${pagination.page}&page_size=${pagination.pageSize}"
     client.GET[Paged[Event]](url)
   }
 
--- a/app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
+++ b/app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
@@ -76,17 +76,14 @@ class BindingTariffClassificationConnector @Inject()(appConfig: AppConfig, clien
   }
 
   def findEvents(reference: String, pagination: Pagination)(implicit hc: HeaderCarrier): Future[Paged[Event]] = {
-    findFilteredEvents(reference,pagination, None)
+    findFilteredEvents(reference,pagination, Set.empty)
   }
 
-  def findFilteredEvents(reference: String, pagination: Pagination, onlyEventTypes: Option[Set[EventType]])(implicit hc: HeaderCarrier): Future[Paged[Event]] = {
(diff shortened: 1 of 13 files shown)


app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
```diff
@@ -75,7 +76,17 @@ class BindingTariffClassificationConnector @Inject()(appConfig: AppConfig, clien
   }
 
   def findEvents(reference: String, pagination: Pagination)(implicit hc: HeaderCarrier): Future[Paged[Event]] = {
-    val url = s"${appConfig.bindingTariffClassificationUrl}/cases/$reference/events?page=${pagination.page}&page_size=${pagination.pageSize}"
+    findFilteredEvents(reference,pagination, None)
+  }
+
+  def findFilteredEvents(reference: String, pagination: Pagination, onlyEventTypes: Option[Set[EventType]])(implicit hc: HeaderCarrier): Future[Paged[Event]] = {
```
pllaurado: We could do this **onlyEventTypes** without Option . We could use empty set  instead none.

app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
```diff
@@ -75,7 +76,17 @@ class BindingTariffClassificationConnector @Inject()(appConfig: AppConfig, clien
...
+
+    val searchParam = onlyEventTypes match {
+      case Some(only) => s"case_reference=$reference${only.mkString("&type=","&type=","&")}"
```
pllaurado: only means eventsType ?  :-)

pllaurado: I guess case_reference=$reference can be extract it  to use it one 
1st -> returnValue = case_reference=$reference 
2nd -> returnValue & onlyEventTypes.map(o =>  s"&type=$o").mkString("")  

just an idea

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/CaseController.scala
```diff
@@ -190,7 +195,7 @@ class CaseController @Inject()(verify: RequestActions,
   private def showActivity(c: Case, f: Form[ActivityFormData])
                           (implicit request: AuthenticatedRequest[AnyContent]): Future[HtmlFormat.Appendable] = {
     for {
-      events <- eventsService.getEvents(c.reference, NoPagination())
+      events <- eventsService.getFilteredEvents(c.reference, NoPagination(),Some(EventType.values.diff(Set(EventType.SAMPLE_STATUS_CHANGE))))
```
pllaurado: I would prefer to define the positive one which means define all the ones you want. But quite clever any way ;-)

app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
```diff
@@ -75,7 +76,17 @@ class BindingTariffClassificationConnector @Inject()(appConfig: AppConfig, clien
...
+      case None => s"case_reference=$reference&"
```
pllaurado: probably this last & should be on the rest call . easy peasy

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/CaseController.scala
aaroncarroll82: I did it like that to start with and maybe it is clearer actually, but I thought this might be better as firstly it makes the line really long if all are listed, and secondly I thought for future safety, if new types are added they wouldn't automatically be shown on activity feed unless they were added here, whereas this way keeps it futureproof as it's always 'everything except sample status'

app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
aaroncarroll82: Simplified this based on comment, clearer and shorter now I think. Thanks

app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
aaroncarroll82: I've changed to use empty set and it's a bit simpler

app/uk/gov/hmrc/tariffclassificationfrontend/connector/BindingTariffClassificationConnector.scala
aaroncarroll82: Changed this alongside simplification changes above.

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/CaseController.scala
pllaurado: I guess we will have the problem either way. In this case if we add something we don't want in this specific list we will add it for anyway. I prefer positive code but I agree to keep this one..

app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/sample/sample_status_events_table.scala.html
```diff
@@ -0,0 +1,58 @@
...
+        </tr>
+    </thead>
+    <tbody>
+    @for((event: Event, index: Int) <- events.results.zipWithIndex) {
+        <tr id="sample-status-events-row-@{index}">
+            <td id="sample-status-events-row-@{index}-operator" class="sample-status-cell">@event.operator.name.getOrElse("Unknown")</td>
+
+            <td id="sample-status-events-row-@{index}-content" class="sample-status-cell-desc">
+                @{
+                    val body = event.details.asInstanceOf[SampleStatusChange]
```
pllaurado: Not entirely sure if we should do it in this way.  The assumption that this events always will be a **SampleStatusChange** is a bit dangerous. Could we protected it with an **if**  ? maybe if you get a different one you would like to error or just pass

aaroncarroll82: I'll add a safety check around this, in same style as the table on activity view. This will throw a specific exception for unhandled types.

PR #207, opened 2019-05-20
Dit 1011 - ADR Appeal Status

SteveSugden:


app/uk/gov/hmrc/tariffclassificationfrontend/models/AppealType.scala
```diff
@@ -18,17 +18,20 @@ package uk.gov.hmrc.tariffclassificationfrontend.models
...
+    case _ => "Appeal"
+  }
+
   type AppealType = Value
-  val REVIEW  = Value(1)
-  val APPEAL_TIER_1 = Value(2)
-  val APPEAL_TIER_2 = Value(3)
-  val COURT_OF_APPEALS = Value(4)
-  val SUPREME_COURT = Value(5)
+  val REVIEW, APPEAL_TIER_1, APPEAL_TIER_2, COURT_OF_APPEALS, SUPREME_COURT, ADR = Value
```
aaroncarroll82: I know it probably doesn't matter but ADR is the first enum value in the backend, but last value here in front end. Should it be consistent?

app/uk/gov/hmrc/tariffclassificationfrontend/views/appeal_choose_type.scala.html
```diff
@@ -35,6 +35,7 @@ <h2 id="appeal_choose_type-heading" class="heading-large">Choose review or a lev
    @input_radiogroup(
     field = form("type"),
     options = Seq(
+     AppealType.ADR,
```
aaroncarroll82: Linked with the ordering of enum vals in comment above. If they were in 'correct' order in the enum we could use them here with an AppealType.values.toList.ordered type call, rather than refer to them all specifically

SteveSugden: Probably, yes.   I just added the new one in the position required.   I guess its easier to change this way if the designers change their mind.

app/uk/gov/hmrc/tariffclassificationfrontend/models/AppealType.scala
SteveSugden: I'll change it.  Sure it doesn't matter though.

PR #208, opened 2019-05-20
Dit 939

aaroncarroll82:


retroredge:
The solution will address the DAC report request however the general advice on the Web seems to be that "Using a tabindex greater than 0 is considered an anti-pattern".

aaroncarroll82:
> Using a tabindex greater than 0 is considered an anti-pattern

I'm open to suggestions on how else to achieve!

PR #211, opened 2019-05-22
Dit 1034 - custom page titles

aaroncarroll82:
Custom page titles for each distinct page in the UI.

Probably don't need any unit tests since the page titles are checked in acceptance tests and it would just be repeating.

### DIFF
--- a/app/uk/gov/hmrc/tariffclassificationfrontend/views/includes/main.scala.html
+++ b/app/uk/gov/hmrc/tariffclassificationfrontend/views/includes/main.scala.html
@@ -20,7 +20,8 @@
   contentHeader: Option[Html] = None,
   bodyClasses: Option[String] = None,
   mainClass: Option[String] = None,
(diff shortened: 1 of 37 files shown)


app/uk/gov/hmrc/tariffclassificationfrontend/views/includes/main.scala.html
```diff
@@ -35,7 +36,7 @@
 
 
 @govuk_wrapper(appConfig = appConfig,
-               title = "Tariff Classification",
+               title = s"${customTitle.getOrElse("Tariff classification")}",
```
pllaurado: Why is this s"{}" ? customTitle.getOrElse("Tariff classification")

aaroncarroll82: Good point actually - because I had that line as something more complicated before where it needed that, can be simplified now

PR #212, opened 2019-05-23
DIT-945 Introduce post redirect get pattern

pllaurado:


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/SuspendCaseController.scala
```diff
@@ -41,24 +41,34 @@ class SuspendCaseController @Inject()(verify: RequestActions,
   override protected val caseService: CasesService = casesService
   private val form: Form[Boolean] = MandatoryBooleanForm.form("suspend_case")
 
-  def suspendCase(reference: String): Action[AnyContent] = (verify.authenticated andThen verify.casePermissions(reference) andThen verify.mustHave(Permission.SUSPEND_CASE)).async { implicit request =>
+  def getSuspendCase(reference: String): Action[AnyContent] = (verify.authenticated andThen verify.casePermissions(reference) andThen verify.mustHave(Permission.SUSPEND_CASE)).async { implicit request =>
```
SteveSugden: I like the name changes - so we can tell what action the method is handling

aaroncarroll82:
> Looks like Aaron was right (not about the ";"). Changes look OK - v. big 2

I'm still missing Java obviously!

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/RenderCaseAction.scala
```diff
@@ -47,13 +47,35 @@ trait RenderCaseAction extends FrontendController with I18nSupport {
     }
   }
 
+  protected def defaultRedirect (implicit request: AuthenticatedCaseRequest[_]): Future[Result] = {
+    successful(Redirect(redirect(request.`case`.reference)))
+  }
```
aaroncarroll82: Very minor - can you add a new line here so the two methods don't touch

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/RenderCaseAction.scala
```diff
@@ -47,13 +47,35 @@ trait RenderCaseAction extends FrontendController with I18nSupport {
...
+  protected def validateAndRedirect(toHtml: Case => Future[Call])
+                                     (implicit request: AuthenticatedCaseRequest[_]): Future[Result] = {
+
+    if (isValidCase(request.`case`)(request)) {
+      toHtml(request.`case`).map(Redirect)
```
aaroncarroll82: Is it maybe possible to refactor slightly and create a default implementation for this call along the lines of the defaultRedirect above? There's 3 places in the file where we make a call of 'convert case to html and then map to x'

pllaurado: I would agree is was just one parameter, but in this case to refactor would be send a request or Case and the action after the map. Not sure if will look more clear ... I think it looks quite readable now :)

PR #213, opened 2019-05-23
DIT-1035 Text input markup fixes (DAC)

retroredge:


app/uk/gov/hmrc/tariffclassificationfrontend/views/forms/components/input_textarea.scala.html
```diff
@@ -30,23 +30,26 @@
...
+    @field.errors.map { error =>
+        <span class="error-message" id="error-message-@{field.id}-input">@customErrorMessage.getOrElse( messages(error.message, error.args: _*))</span>
+    }
+
     <textarea
             class="form-control w-100 @inputClass"
             id="@{field.id}"
             name="@{field.id}"
-            aria-describedby="error-message-@{field.id}-input"
+            aria-describedby="@{field.id}-hint"
```
aaroncarroll82: Looking at the guidelines, this should be something like
aria-describedby="@{field.id}-hint error-message-@{field.id}-input" in case of error being present (so referencing both normal hint and error).

PR #215, opened 2019-05-28
DIT-1047 - Remove ambiguities from advanced search images alt

aaroncarroll82:


aaroncarroll82:
AT did need an update. Done that and raised a separate PR.

PR #218, opened 2019-05-29
DIT-1020 add end-point for getting (viewing) attachment by file id

SteveSugden:


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/ViewAttachmentController.scala
```diff
@@ -0,0 +1,46 @@
...
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+@Singleton
+class ViewAttachmentController @Inject()(verify: RequestActions,
+                                         fileService: FileStoreService,
+                                         val messagesApi: MessagesApi,
+                                         implicit val appConfig: AppConfig) extends FrontendController with I18nSupport {
+
+  def get(id: String): Action[AnyContent] = (verify.authenticated andThen verify.mustHave(Permission.VIEW_CASES)).async
```
pllaurado: Should we have view_attachments permission?

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/ViewAttachmentController.scala
```diff
@@ -0,0 +1,46 @@
...
+  { implicit request: Request[AnyContent] =>
+    fileService.getFileMetadata(id) map {
```
pllaurado: I would  add **case _**  in case you get any unexpected answer
Just as a preference  
      case Some(file) if isFileReady(file) => Redirect(Call(method = "GET", url = fileSubmitted.url.get))
      case Some(file) => Ok(views.html.view_attachment_unavailable(file))
      case _ => unexpected_failure_view

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/ViewAttachmentController.scala
ed0906: Read Only can view attachments cant they?

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/ViewAttachmentController.scala
```diff
@@ -0,0 +1,46 @@
...
+      case Some(fileSubmitted: FileMetadata) if fileSubmitted.url.isDefined && fileSubmitted.scanStatus.contains(ScanStatus.READY) =>
```
ed0906: I think these days `fileSubmitted.url.isDefined` is a sufficient check. Filestore doesn't send out unsafe urls

app/uk/gov/hmrc/tariffclassificationfrontend/views/view_attachment_unavailable.scala.html
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.tariffclassificationfrontend.config.AppConfig
+@import uk.gov.hmrc.tariffclassificationfrontend.views.html.includes.main
+@import uk.gov.hmrc.tariffclassificationfrontend.models.response.{ScanStatus, FileMetadata}
+
+@(fileSubmitted: Option[FileMetadata])(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
+
+@main(bodyClasses = None, customTitle = Some(messages("view.attachment.unavailabe.title"))) {
```
ed0906: `unavailabe` should be `unavailable`

conf/messages
```diff
@@ -281,4 +281,11 @@ page.title.choose.appeal.status=Choose appeal status
 page.title.choose.appeal.type=Choose appeal type
 page.title.extended.use=Extended use of this ruling
 page.title.ruling.not.found=Ruling not found
-page.title.case.queues=Tariff classification case queues
\ No newline at end of file
+page.title.case.queues=Tariff classification case queues
+
+
+view.attachment.unavailabe.title=Attachment unavailable
```
ed0906: `unavailabe` should be `unavailable`

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/ViewAttachmentController.scala
SteveSugden: RO users can.  I wondered about a new permission.  But the existing way of viewing an attachment does not require any special permission (other than view case) so I thought that would be ok.

ed0906:
Maybe an IT test?

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/ViewAttachmentController.scala
SteveSugden: What would the "unexpected failure view" show other than the existing "something has gone wrong" error that you would get when the match fails?

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/ViewAttachmentController.scala
SteveSugden: So you're saying that if the url exists then ScanStatus will always be ready?

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/ViewAttachmentController.scala
pllaurado: Ok, then VIEW_CASES implicitly means VIEW_ATTACHMENTS :)

PR #220, opened 2019-05-29
DIT-1079 Error validation on tabs

pllaurado:


app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/error_summary.scala.html
```diff
@@ -0,0 +1,32 @@
...
+@if(errors.nonEmpty) {
+    <div class="error-summary error-summary--show" id="error-summary" role="group" aria-labelledby="error-summary-heading" tabindex="@startAtTabIndex.map(index => index + 1)">
+
+        <h2 class="heading-medium error-summary-heading" id="error-summary-heading">
+            @messages("error.summary.title")
+        </h2>
+
+        <ul role="list" class="error-summary-list">
+            @for((error: FormError, index: Int) <- errors.zipWithIndex) {
+                <li><a href="#@{error.key}" tabindex="@startAtTabIndex.map(x => x+2+(0*index))">@messages(error.message, error.args:_*)</a></li>
```
aaroncarroll82: 0*index in last part of this will always yield 0. Might not cause an issue as you can have repeated tab indexes, but if it's not needed might as well remove it, if it is needed, should produce unique results.

ed0906: Agreed this will always map to `x+2`

pllaurado: are u suggesting  then **x+ 2 + index** ?

PR #221, opened 2019-05-30
Dit 1084 - Spike minimum changes required for liabilities

SteveSugden:


app/uk/gov/hmrc/tariffclassificationfrontend/views/templates/ruling_template.scala.html
```diff
@@ -97,21 +99,21 @@ <h2>@messages("pdf.ruling.section.holder.heading")</h2>
                 <table>
                     <tr>
                         <td class="section-label">@messages("pdf.ruling.section.holder.name.label")</td>
-                        <td id="ruling.holder.businessName">@c.application.holder.businessName</td>
+                        <td id="ruling.holder.businessName">@btiApp.holder.businessName</td>
```
ed0906: Can this use your new business name method?

SteveSugden: Can do.  But this is definitely a BTI page

PR #224, opened 2019-05-30
DIT-1023 - Cancel ruling with critical evidence

aaroncarroll82:
Draft PR for now as I haven't even looked at unit tests yet


app/uk/gov/hmrc/tariffclassificationfrontend/models/CancelRuling.scala
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.tariffclassificationfrontend.models
+
+import uk.gov.hmrc.tariffclassificationfrontend.models.CancelReason.CancelReason
+
+case class CancelRuling(cancelReason: String, note: String)
```
ed0906: maybe `RulingCancellation`?

app/uk/gov/hmrc/tariffclassificationfrontend/service/CasesService.scala
```diff
@@ -35,6 +35,7 @@ import uk.gov.hmrc.tariffclassificationfrontend.models.request.NewEventRequest
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
+//noinspection ScalaStyle
```
ed0906: Why are these appearing?

aaroncarroll82: They won't be there in final version.

app/uk/gov/hmrc/tariffclassificationfrontend/models/CancelRuling.scala
aaroncarroll82: Yep think that is a better name, will use that.

PR #226, opened 2019-05-31
DIT 1026 - Render Notes & Reason codes in the Case Activity Log

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/activity/activity_case_cancellation.scala.html
```diff
@@ -0,0 +1,49 @@
...
+
+@import uk.gov.hmrc.tariffclassificationfrontend.config.AppConfig
+@import uk.gov.hmrc.tariffclassificationfrontend.models.request.AuthenticatedRequest
+@import uk.gov.hmrc.tariffclassificationfrontend.models.{CancelReason, CancellationCaseStatusChange, Event}
+@import uk.gov.hmrc.tariffclassificationfrontend.views.html.partials.progressive_disclosure
+@(index: Int, event: Event)(implicit request: AuthenticatedRequest[_], messages: Messages, appConfig: AppConfig)
+
+@content(body: CancellationCaseStatusChange) = {
+ <p>
+  <span class="bold">Reason:</span>
```
aaroncarroll82: Is the content right here? Can't see anywhere on the prototypes/designs where we should show **Reason**: - it seems to be like that for note, but the other stuff seems to be styled just the reason itself in bold, e.g **Invalidated due to EU measure (62)**

ed0906: I couldn't find any screenshots for it, let's see if they say anything at post amigos

PR #231, opened 2019-06-05
DIT-1025 - sample changes for referred case for sample

SteveSugden:


app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/sample/sample_status_events_table.scala.html
```diff
@@ -44,7 +44,16 @@ <h3 class="heading-medium" id="sample-status-events-heading">Sample activity</h3
                             val body = event.details.asInstanceOf[SampleStatusChange]
                             progressive_disclosure(
                                 id = s"sample-status-events-row-$index",
-                                summary = SampleStatus.format(body.to),
+                                summary = s"Sample status changed from ${SampleStatus.format(body.from, false)} to ${SampleStatus.format(body.to, false)}",
+                                body = body.comment.map(Html(_))
```
aaroncarroll82: It didn't support it at the time of you putting this in - but progressive disclosure can now support tabindexes, so we should probably set them here to allow for expanding these elements using keyboard nav.

SteveSugden: Good spot

conf/app.routes
```diff
@@ -87,6 +87,9 @@ POST        /cases/:reference/change-appeal/:appealId           @uk.gov.hmrc.tar
 GET         /cases/:reference/sample/status                     @uk.gov.hmrc.tariffclassificationfrontend.controllers.SampleController.chooseStatus(reference: String)
 POST        /cases/:reference/sample/status                     @uk.gov.hmrc.tariffclassificationfrontend.controllers.SampleController.updateStatus(reference: String)
 
+GET         /cases/:reference/sample/returnStatus               @uk.gov.hmrc.tariffclassificationfrontend.controllers.SampleReturnController.chooseStatus(reference: String)
+POST        /cases/:reference/sample/returnStatus               @uk.gov.hmrc.tariffclassificationfrontend.controllers.SampleReturnController.updateStatus(reference: String)
```
aaroncarroll82: Should these endpoints have camelcase in? Looks like we typically hyphenate when multi word. Thought there was possibly some kind of accessibility concern around camel cased URLs?

app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/sample/sample_status_events_table.scala.html
SteveSugden: Fixed.   But the progressive bit doesn't appear to be used 'cos there is no 'comment'.  But done anyway

conf/app.routes
SteveSugden: Another good spot.  Fixed

SteveSugden: in my defence, I got confused with the camelCase variable names (not same thing I know)

app/uk/gov/hmrc/tariffclassificationfrontend/models/SampleReturn.scala
```diff
@@ -0,0 +1,32 @@
...
+  type SampleReturn = Value
+
+  val YES, NO, TO_BE_CONFIRMED = Value
+
+  def format(status: Option[SampleReturn]): String = {
+    status match {
+      case Some(YES) => "Yes"
+      case Some(NO) => "No"
+      case Some(TO_BE_CONFIRMED) => "To be confirmed"
+      case _ => "Unknown"
```
aaroncarroll82: Was just wondering with this - should case of None just be 'To be confirmed'? Still makes sense, but possibly looks  bit more pleasant than Unknown

SteveSugden: Should never happen.  If the status is None then it shouldn't be rendered.

pllaurado: should never happen = thrown exception ?

app/uk/gov/hmrc/tariffclassificationfrontend/models/SampleReturn.scala
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package uk.gov.hmrc.tariffclassificationfrontend.models
+
+object SampleReturn extends Enumeration {
+  type SampleReturn = Value
+
+  val YES, NO, TO_BE_CONFIRMED = Value
+
+  def format(status: Option[SampleReturn]): String = {
```
pllaurado: I prefer to use Options just to return or store when something can be empty, but create a function with option param ... is just an opinion

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/SampleReturnController.scala
```diff
@@ -0,0 +1,74 @@
...
+                                       override val messagesApi: MessagesApi,
+                                       override implicit val config: AppConfig) extends StatusChangeAction[Option[SampleReturn]] {
+
+  override protected val requiredPermission: Permission.Value = Permission.EDIT_SAMPLE
+
+  override protected def redirect: String => Call = routes.CaseController.trader
+
+  override protected def isValidCase(c: Case)(implicit request: AuthenticatedRequest[_]): Boolean = {
+    true //No constraints on when the case is valid
+  }
```
ed0906: Minor but the braces arent required

app/uk/gov/hmrc/tariffclassificationfrontend/models/Sample.scala
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package uk.gov.hmrc.tariffclassificationfrontend.models
+
+import uk.gov.hmrc.tariffclassificationfrontend.models.SampleReturn.SampleReturn
+import uk.gov.hmrc.tariffclassificationfrontend.models.SampleStatus.SampleStatus
+
+case class Sample
+(
+  status: Option[SampleStatus] = None,
```
ed0906: Option? how come?  Isnt `Sample` Optional on the Case level?

ed0906: I take it back, it  isnt

PR #232, opened 2019-06-05
1102 - Issues with errors

aaroncarroll82:
* Remove extra error_summary class. 
* Order errors correctly on refer case screen. 
* Enable possibility to click on errors in summary and jump to errors in the embedded parts of radio group
* Make error descriptions match in note field and error summary
* Make error description match for cancel reason and error summary
* Make sure note field is constructed in accessible way on the crit evidence pages


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/SuspendCaseController.scala
```diff
@@ -41,7 +41,7 @@ class SuspendCaseController @Inject()(verify: RequestActions,
 
   override protected val config: AppConfig = appConfig
   override protected val caseService: CasesService = casesService
-  private val form: Form[String] = AddNoteForm.form
+  private val form: Form[String] = AddNoteForm.getForm("error.empty.suspend.note")
```
ed0906: Seems a bit of a strange thing to be passing into a `getForm` method? maybe just pass in `suspend` and the `error.empty.{x}.note` key is generated inside the form?

app/uk/gov/hmrc/tariffclassificationfrontend/forms/mappings/FormMappings.scala
```diff
@@ -46,4 +46,17 @@ object FormMappings {
         Map(key -> value)
     })
 
+  def oneOf(errorKey: String = "error.required", enumeration: Enumeration): FieldMapping[String] =
```
ed0906: Is this unit tested?

aaroncarroll82: Not directly (none of the methods in FormMappings are), but it is as part of the form that uses it. Couldn't hurt to add some unit tests for the methods in here too I guess

aaroncarroll82: Added unit tests for all 3 methods now.

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/SuspendCaseController.scala
aaroncarroll82: updated as suggested

PR #238, opened 2019-06-10
DIT 1083 - view a new Liability case 

pllaurado:


app/uk/gov/hmrc/tariffclassificationfrontend/views/case_details.scala.html
```diff
@@ -28,11 +29,15 @@
     @case_heading(c, displayBackLink = true)
 
     <div class="tabs">
-        @case_nav(page, c)
-
```
pllaurado: Because now we will have a BTI or Liability this will render different tabs. Now we need to double check all the end points that are only available for BTI's.

PR #243, opened 2019-06-11
DIT 1043 Edit Liability view

pllaurado:


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/LiabilityController.scala
```diff
@@ -35,23 +39,103 @@ import scala.concurrent.Future.successful
...
+        )
+    }
+  }
+
+  private def toLiabilityForm(l: LiabilityOrder): Form[LiabilityFormData] = {
+    LiabilityFormData.form.fill(
+      LiabilityFormData(
+        entryDate = l.entryDate,
+        traderName = l.traderName,
+        goodName = l.goodName.getOrElse(""),
```
SteveSugden: Can these form fields be Option[String]?  then you wouldn't have to do getOrElse

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/LiabilityController.scala
```diff
@@ -35,23 +39,103 @@ import scala.concurrent.Future.successful
...
+      case c: Case =>
+        successful(
+          Ok(
+            liability_details_edit(c, toLiabilityForm(c.application.asLiabilityOrder), Some(tabIndexFor(LIABILITY)))
+          )
+        )
+    }
+  }
+
+  private def toLiabilityForm(l: LiabilityOrder): Form[LiabilityFormData] = {
```
ed0906: Can we try and do this with the form mapping. See `DecisionFormData` with the Liability form

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/LiabilityController.scala
```diff
@@ -35,23 +39,103 @@ import scala.concurrent.Future.successful
...
+        traderCommodityCode = l.traderCommodityCode.getOrElse(""),
+        officerCommodityCode = l.officerCommodityCode.getOrElse(""),
+        contactName = l.contact.name,
+        contactEmail = Some(l.contact.email),
+        contactPhone = l.contact.phone.getOrElse("")
+      )
+    )
+  }
+
+  def mergeLiabilityIntoCase(c: Case, validForm: LiabilityFormData): Case = {
```
ed0906: As above, can we try and  do this with the form mapping

app/uk/gov/hmrc/tariffclassificationfrontend/forms/LiabilityFormData.scala
```diff
@@ -0,0 +1,83 @@
...
+  private val instantToTuple: Instant => FormDate = { date =>
+    val offsetDate = date.atOffset(UTC).toLocalDate
+    (offsetDate.getDayOfMonth, offsetDate.getMonthValue, offsetDate.getYear)
+  }
+
+  private val validDateFormat: FormDate => Boolean = {
+    case (day, month, year) => Try(LocalDate.of(year, month, day)).isSuccess
+  }
+
+  private val emailRegex = """^[a-zA-Z0-9\.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$""".r
```
SteveSugden: Do we really have to validate email like this?  email regex is notoriously hard to get "right"

SteveSugden: TraderUI appears to be using org.apache.commons.validator.routines.EmailValidator.  Maybe we could do the same here?

ed0906:
Looks like theres conflicts

app/uk/gov/hmrc/tariffclassificationfrontend/forms/LiabilityFormData.scala
ed0906: I think he copied this from the Scala Play email validation constraint

pllaurado: Is either to introduce a new library and keep it updated just for this email validation... There is another Validation on play which checks the email in the same way you would check a text, number, etc... but because an empty is valid and we need our own error message we have created in this way.

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/LiabilityController.scala
ed0906: Done

PR #253, opened 2019-06-13
BAU - fix more clickable errors. Content change to make LIVE tag look…

aaroncarroll82:
… nicer on case

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -314,7 +314,9 @@ table tbody td[align="right"], table tbody th[align="right"], table thead th[ali
 }
 
 #case-status-section .live-red-text {
-  padding-right: 10px;
+  padding-right: 8px;
+  font-size: 16px;
+  font-family: Arial;
 }
 
 .button-link {
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -313,6 +313,10 @@ table tbody td[align="right"], table tbody th[align="right"], table thead th[ali
   font-weight: 600;
 }
 
+#case-status-section .live-red-text {
+  padding-right: 10px;
+}
+
 .button-link {
   background: none !important;
   color: inherit;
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/statuses/case_status.scala.html
```diff
@@ -19,6 +19,10 @@
 @import uk.gov.hmrc.tariffclassificationfrontend.models.{CancelReason, Case, CaseStatus}
 @(c: Case, id: String, suppressLiveIndicator: Boolean = false)(implicit appConfig: AppConfig)
 
+@if(c.application.isLiveLiabilityOrder && !suppressLiveIndicator) {
```
ed0906: Presume this means that it comes first now?

app/assets/stylesheets/layout.scss
```diff
@@ -313,6 +313,10 @@ table tbody td[align="right"], table tbody th[align="right"], table thead th[ali
   font-weight: 600;
 }
 
+#case-status-section .live-red-text {
+  padding-right: 10px;
```
ed0906: Looks like the font size is different to the lozenge, we should probably be consistent

app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/statuses/case_status.scala.html
aaroncarroll82: yep

app/assets/stylesheets/layout.scss
aaroncarroll82: it's a bit of a weird one actually, if you put it to same font-size it somehow looks smaller. I think it's an optical illusion because of the lozenge background surround.

aaroncarroll82: Updated to be the same and it looks a little bit better I think

ed0906: Yeh I tried it too, strange

PR #255, opened 2019-06-13
Dit 1094 - liability samples

SteveSugden:


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/StatusChangeAction.scala
```diff
@@ -35,29 +35,29 @@ trait StatusChangeAction[T] extends RenderCaseAction {
...
+  protected def chooseStatusView(c: Case, preFilledForm: Form[T], options: Option[String] = None)(implicit request: Request[_]): Html
 
   protected def update(c: Case, status: T, operator: Operator)(implicit hc: HeaderCarrier): Future[Case]
 
   protected def onSuccessRedirect(reference: String): Call
 
   protected val requiredPermission: Permission
 
-  def chooseStatus(reference: String): Action[AnyContent] = (verify.authenticated andThen verify.casePermissions(reference) andThen
+  def chooseStatus(reference: String, options: Option[String] = None): Action[AnyContent] = (verify.authenticated andThen verify.casePermissions(reference) andThen
```
aaroncarroll82: What're the options for here? They don't seem to be used. Unlike for the chooseStatusView method, all classes that use this trait override this method, and none of them use the options. Maybe an idea that you started off with and ended up not using?

app/uk/gov/hmrc/tariffclassificationfrontend/models/SampleStatus.scala
```diff
@@ -18,11 +18,12 @@ package uk.gov.hmrc.tariffclassificationfrontend.models
 
 object SampleStatus extends Enumeration {
   type SampleStatus = Value
-  val AWAITING, MOVED_TO_ACT, MOVED_TO_ELM, SENT_FOR_ANALYSIS, SENT_TO_APPEALS, STORAGE, RETURNED_APPLICANT,
+  val NONE, AWAITING, MOVED_TO_ACT, MOVED_TO_ELM, SENT_FOR_ANALYSIS, SENT_TO_APPEALS, STORAGE, RETURNED_APPLICANT,
```
aaroncarroll82: What does this give us having an option called NONE, over just setting it to an Option value of None? Far as I can see everywhere it's used in a model object it's optional.

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/StatusChangeAction.scala
SteveSugden: ExtendedUseCase controller appears to use the trait method.   I wanted it in the SampleController so that I could use it to determine what form to render.

app/uk/gov/hmrc/tariffclassificationfrontend/models/SampleStatus.scala
SteveSugden: It's needed because we have chosen to re-purpose `None` as meaning "liability sample not being sent".  Now `Some(NONE)` means "no sample status and Some(_) means "liability sample is being sent (for a liability case).

Hope that makes sense.

aaroncarroll82: Ahh OK got you - couldn't get that from my brief look, but fair enough!

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/StatusChangeAction.scala
aaroncarroll82: Yep, you're right, sorry, didn't notice that the extendedusecase class fell back to the default method in trait

app/uk/gov/hmrc/tariffclassificationfrontend/models/SampleStatus.scala
```diff
@@ -18,11 +18,12 @@ package uk.gov.hmrc.tariffclassificationfrontend.models
...
   RETURNED_PORT_OFFICER, RETURNED_COURIER, DESTROYED = Value
 
   def format(status: Option[SampleStatus], initialCaps: Boolean = true): String = {
     val text = status match {
+      case Some(NONE) => "none"
```
ed0906: You can have this as the `_` case

SteveSugden: true

ed0906: as in, you can remove this line or I think do `case _ | "none"`

app/uk/gov/hmrc/tariffclassificationfrontend/views/forms/liability_sending_sample_form.scala.html
```diff
@@ -0,0 +1,40 @@
...
+@import uk.gov.hmrc.tariffclassificationfrontend.views.forms.components.RadioOption
+@import uk.gov.hmrc.tariffclassificationfrontend.views.html.forms.components.input_radiogroup
+@(c: Case, field: Field)(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
+
+
+@helper.form(action = routes.SampleController.updateStatus(c.reference, Some("liability")), 'id -> "sample_status_form") {
+    @helper.CSRF.formField
+    @input_radiogroup(
+        field = field,
+        options = Seq(RadioOption(SampleStatus.AWAITING.toString, "Yes"), RadioOption("", "No")),
```
ed0906: Is your No option not `SampleStatus.NONE`?

SteveSugden: No.  Yes is Some(Status) (inc. Some(NONE)).   No is None which is achieved via the radio option value "".
I chose AWAITING as a sensible initial sample status when the "is sending sample" state is first turned on.

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/SampleController.scala
```diff
@@ -54,16 +54,19 @@ class SampleController @Inject()(override val verify: RequestActions,
 
   override protected def status(c: Case): Option[SampleStatus] = c.sample.status
 
-  override protected def chooseStatusView(c: Case, notFilledForm: Form[Option[SampleStatus]])
+  protected def chooseStatusView(c: Case, notFilledForm: Form[Option[SampleStatus]], options: Option[String])
                                          (implicit request: Request[_]): Html = {
-    views.html.change_sample_status(c, notFilledForm)
+    if(options.contains("liability"))
```
ed0906: Why  do we need options when we can just do `if(c.application.isLiability)`?

SteveSugden: Because liabilities can also set sample status (i.e. need to show the original form).

app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/sample/sample_details_bti.scala.html
```diff
@@ -0,0 +1,78 @@
...
+        </div>
+        <div id="sample-requested-return" class="grid-row">
+            <dt class="column-one-third">
+            @messages("case.sample.request.return")
+            </dt>
+            <dd id="sample-requested-return-value" class="column-one-third">
+                @SampleReturn.format(c.sample.returnStatus)
+            </dd>
+            <dd class="column-one-third">
+                <a id="sample-requested-return-change" tabindex="@startAtTabIndex" href="sample/return-status">@messages("case.sample.request.change")</a>
```
ed0906: Can we use routes references in the href please :)

app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/sample/sample_details_liability.scala.html
```diff
@@ -0,0 +1,56 @@
...
+
+    <div id="liability-sending-samples" class="grid-row">
+        <dt class="column-one-third">
+        @messages("case.liability.provide-samples")
+        </dt>
+        <dd class="column-one-third">
+        @if(c.sampleToBeProvided){ @messages("answer.yes") } else { @messages("answer.no")}
+        </dd>
+        <dd class="column-one-third">
+        <a id="sample-return-change" tabindex="@startAtTabIndex" href="sample/status?options=liability">@messages("case.sample.request.change")</a>
```
ed0906: Can we use routes references in the href please :)

app/uk/gov/hmrc/tariffclassificationfrontend/views/partials/sample/sample_details_liability.scala.html
```diff
@@ -0,0 +1,56 @@
...
+    @if(c.sampleToBeProvided){
+        <div id="liability-returning-samples" class="grid-row">
+            <dt class="column-one-third">
+                @messages("case.liability.returning-samples")
+            </dt>
+            <dd class="column-one-third">
+                @if(c.sampleToBeReturned){ @messages("answer.yes") } else { @messages("answer.no")}
+            </dd>
+            <dd class="column-one-third">
+                <a id="sample-requested-return-change" tabindex="@startAtTabIndex" href="sample/return-status">@messages("case.sample.request.change")</a>
```
ed0906: Can we use routes references in the href please :)

app/uk/gov/hmrc/tariffclassificationfrontend/views/forms/liability_sending_sample_form.scala.html
```diff
@@ -0,0 +1,40 @@
...
+        id = "sample_status_form",
+        preSelect = false,
+        label = Some(messages("case.liability.sending.sample.change-return-header")),
+        labelClass = Some ("heading-large")
+    )
+    <div>
```
ed0906: Theres a `confirm_or_cancel_buttons` components you can use here

PR #256, opened 2019-06-14
DIT-1107 - Liability Ruling PDF view

aaroncarroll82:


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/PdfDownloadController.scala
```diff
@@ -47,6 +47,14 @@ class PdfDownloadController @Inject()(authenticatedAction: AuthenticatedAction,
     }
   }
 
+  def liabilityPdf(reference: String): Action[AnyContent] = authenticatedAction.async { implicit request =>
+    caseService.getOne(reference) flatMap {
+      case Some(c: Case) if c.decision.isDefined => generatePdf(decision_template(c, c.decision.get), s"LiabilityDecision_$reference.pdf")
```
SteveSugden: worth checking that c.application.isLiability?

test/unit/uk/gov/hmrc/tariffclassificationfrontend/views/partials/liabilities/LiabilityDetailsViewSpec.scala
```diff
@@ -347,4 +347,44 @@ class LiabilityDetailsViewSpec extends ViewSpec with MockitoSugar {
 
   }
 
+  "Render view PDF link" when {
+  for(status: CaseStatus <- CaseStatus.values.filter((x => x == CaseStatus.COMPLETED  || x == CaseStatus.CANCELLED))) {
```
SteveSugden: why not just create a Seq(COMPLETED,CANCELLED) here.  You know what you want.

test/unit/uk/gov/hmrc/tariffclassificationfrontend/views/partials/liabilities/LiabilityDetailsViewSpec.scala
```diff
@@ -347,4 +347,44 @@ class LiabilityDetailsViewSpec extends ViewSpec with MockitoSugar {
...
+        views.html.partials.liabilities.liability_details(c = c, liabilityForm = LiabilityDetailsForm.liabilityDetailsCompleteForm(l),
+          decisionForm = form.liabilityCompleteForm(d))(requestWithPermissions(), messages, appConfig))
+
+      doc should containElementWithID("liability-ruling-certificate-link")
+    }
+  }
+}
+
+  "Not render view PDF link" when {
+    for(status: CaseStatus <- CaseStatus.values.filterNot(_ == CaseStatus.COMPLETED).filterNot(_ == CaseStatus.CANCELLED)) {
```
SteveSugden: I'd create a set of the two special statuses (and use in above test) then use here CaseStatus.values.filterNot(set.contains(_))

conf/app.routes
```diff
@@ -106,6 +106,7 @@ POST        /new-liability                                      @uk.gov.hmrc.tar
 GET         /cases/:reference/keywords/:keyword                 @uk.gov.hmrc.tariffclassificationfrontend.controllers.CaseController.removeKeyword(reference: String, keyword: String)
 
 GET         /pdf/ruling/:reference                              @uk.gov.hmrc.tariffclassificationfrontend.controllers.PdfDownloadController.rulingPdf(reference: String)
+GET         /pdf/decision/:reference                            @uk.gov.hmrc.tariffclassificationfrontend.controllers.PdfDownloadController.liabilityPdf(reference: String)
```
ed0906: Could we try have just one endpoint with a case type switch? (Combine ruling and decision)

test/unit/uk/gov/hmrc/tariffclassificationfrontend/views/partials/liabilities/LiabilityDetailsViewSpec.scala
aaroncarroll82: True. Have improved it.

test/unit/uk/gov/hmrc/tariffclassificationfrontend/views/partials/liabilities/LiabilityDetailsViewSpec.scala
aaroncarroll82: True. Have improved it.

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/PdfDownloadController.scala
aaroncarroll82: The original endpoint for rulingPdf does not check either - would be better if both did! If I combine the endpoint as per Ed's comment below, then will need to do that anyway, so will all be resolved.

conf/app.routes
aaroncarroll82: I should think we can manage that

PR #257, opened 2019-06-16
BAU - permission fix for edit liability details/decision and sample details

SteveSugden:


app/uk/gov/hmrc/tariffclassificationfrontend/controllers/LiabilityController.scala
```diff
@@ -83,6 +83,11 @@ class LiabilityController @Inject()(verify: RequestActions,
     }
   }
 
+  private def hasPermissionForEdit(c: Case, request: AuthenticatedCaseRequest[AnyContent]) = {
```
aaroncarroll82: I like this over explicitly having the logic stated in each controller method. However, the same condition is also used in the view as well, so possibly worth going one step further an putting into a utility class to sue in both places?

ed0906: I think this is what permissions were originally intended for: 
`Permission.EDIT_LIABILITY` would essentially be `case is new & permitted to create cases OR case is open and assigned`

Instead of making utility classes we could build up the Permission functionality a bit more

ed0906: That approach would mean we could replace all case status checks (and others) with a `hasPermission`

PR #261, opened 2019-06-18
DIT - Permissions Refactor

ed0906:


app/uk/gov/hmrc/tariffclassificationfrontend/models/Permission.scala
```diff
@@ -16,36 +16,273 @@
...
+
+  case object EDIT_SAMPLE extends CasePermission {
+    override def name: String = this.getClass.getName
+    override def appliesTo(`case`: Case, operator: Operator): Boolean = managersOrTeamMembersOnly(operator)
   }
 }
+
+
+
+//object Permission extends Enumeration {
```
ed0906: I'll remove this once I've fixed all the tests

app/uk/gov/hmrc/tariffclassificationfrontend/models/Permission.scala
```diff
@@ -16,36 +16,232 @@
...
+  }
+
+  private def managersOrTeamMembersOnly(operator: Operator): Boolean = operator.role match {
+    case Role.CLASSIFICATION_MANAGER | Role.CLASSIFICATION_OFFICER => true
+    case _ => false
+  }
+
+  private def managersOnly(operator: Operator): Boolean = operator.role == Role.CLASSIFICATION_MANAGER
+
+  private def adminOnly(operator: Operator): Boolean = false
```
aaroncarroll82: This one still to be finalised?

ed0906: We don't have a role for it yet, so currently nobody has this

ed0906: Arguably those roles could be deleted I guess

aaroncarroll82: Personally I think we shouldn't leave stubs like this in if it's not implemented as it'll be confusing for a new team - if they have a task that involves adding a new type of role and supporting it, they'd add it all in at the time. What benefit do we give providing something unused at this stage?

ed0906: Agreed, I'll remove

ed0906:
I'm going to pull out the case status logic in a separate PR

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AttachmentsController.scala
```diff
@@ -110,7 +110,7 @@ class AttachmentsController @Inject()(verify: RequestActions,
   }
 
   def uploadAttachment(reference: String): Action[Either[MaxSizeExceeded, MultipartFormData[TemporaryFile]]] =
-    (verify.authenticated andThen verify.mustHave(Permission.ADD_ATTACHMENT))
+    (verify.authenticated andThen verify.casePermissions(reference) andThen verify.mustHave(Permission.ADD_ATTACHMENT))
```
SteveSugden: Why did you add verify.casePermissions?   Adding adding attachment is not dependant on case ownership

ed0906: ADD_ATTACHMENT is a `CasePermission`  and requires `verify.casePermissions`

ed0906: I could make  it a  `GlobalPermission` but that seems wrong

app/uk/gov/hmrc/tariffclassificationfrontend/models/Permission.scala
```diff
@@ -16,36 +16,232 @@
...
+    override def name: String = nameOf(this)
+    override def appliesTo(`case`: Case, operator: Operator): Boolean = managersOrAssignedTeamMembersOnly(`case`, operator)
+  }
+
+  case object REMOVE_ATTACHMENTS extends CasePermission {
+    override def name: String = nameOf(this)
+    override def appliesTo(`case`: Case, operator: Operator): Boolean = managersOrAssignedTeamMembersOnly(`case`, operator)
+  }
+
+  case object CANCEL_CASE extends CasePermission {
```
SteveSugden: Should this not extend GlobalPermission?  Same with add-not, add-attachment, etc

ed0906: My thoughts were that `GlobalPermission`s  are for when you arent dealing with a single case e.g. Search, Queues etc

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AttachmentsController.scala
SteveSugden: But it is a global permission - i.e. not dependent on ownership

app/uk/gov/hmrc/tariffclassificationfrontend/models/Permission.scala
ed0906: This will also end up with a status check (`COMPLETED`) when I start pulling that out

SteveSugden: ... in which case you need the case (see what I did there).  Fair enough

app/uk/gov/hmrc/tariffclassificationfrontend/controllers/AttachmentsController.scala
ed0906: `CasePermission`s  are specific to restricting things on a single `case`. I'd argue Adding attachments is specific to a single case

app/uk/gov/hmrc/tariffclassificationfrontend/models/Permission.scala
```diff
@@ -16,36 +16,232 @@
...
+    override def name: String = nameOf(this)
+    override def appliesTo(`case`: Case, operator: Operator): Boolean = managersOrAssignedTeamMembersOnly(`case`, operator)
+  }
+
+  case object EDIT_LIABILITY extends CasePermission {
+    override def name: String = nameOf(this)
+    override def appliesTo(`case`: Case, operator: Operator): Boolean = managersOrAssignedTeamMembersOnly(`case`, operator)
+  }
+
+  case object EDIT_RULING extends CasePermission {
```
aaroncarroll82: Just wondering if you were planning to bring in the more complex permissions as part of this change, to make them part of the overall permissions trait? For instance the ones currently held in the LiabilityController along the lines of case is open and operator has permission for edit, **or**, case is new and operator has permission for create

SteveSugden: Stay tuned Aaron...  :-)

ed0906: Yep i'm going to do that as a second PR to keep them small

ed0906: Working on that next

aaroncarroll82: Sorry, the screen didn't refresh, I saw this afterwards -




> PR I'm going to pull out the case status logic in a separate

app/uk/gov/hmrc/tariffclassificationfrontend/models/Permission.scala
```diff
@@ -16,36 +16,218 @@
...
+    EDIT_LIABILITY,
+    EDIT_RULING,
+    APPEAL_CASE,
+    EXTENDED_USE,
+    MOVE_CASE_BACK_TO_QUEUE,
+    EDIT_SAMPLE
+  )
+
+  def from(string: String): Option[Permission] = values.find(_.name == string)
+  def applyingTo(operator: Operator): Set[Permission] = values
```
pllaurado: All together is a great idea (Maybe more roles could mess but is not what this app is intended for)... not much to add here. Maybe I would define lazy structures to not iterate them again and again or some other little changes. But just a preference not a requirement 
`lazy val globalPermission =  values.filter(_.isInstanceOf[GlobalPermission]).map(_.asInstanceOf[GlobalPermission])`
`lazy val casePermissions = values.filter(_.isInstanceOf[CasePermission]).map(_.asInstanceOf[CasePermission])`
`def applyingTo(operator: Operator): Set[Permission] =  globalPermission.filter(_.appliesTo(operator)).map(_.asInstanceOf[Permission])`

PR #262, opened 2019-06-18
DIT-1048 - tab indexing and default selection for advanced search

aaroncarroll82:


app/uk/gov/hmrc/tariffclassificationfrontend/views/components/initial_focus.scala.html
```diff
@@ -14,15 +14,18 @@
...
+            var hasErrors = document.getElementById("error-summary");
+            if (hasErrors) {
+                hasErrors.focus();
+            } else {
+                @if(focusOnThisElement.isDefined){
+                     document.getElementById("@focusOnThisElement").focus();
+                }
+            }
+                }
+    , @withTimeout);
```
aaroncarroll82: This timeout has been introduced because we want to select the component of choice with a slight delay on the search pages because there's some other javascript present that grabs the focus for the keywords box, and we want to do it after that to ensure we win.

PR #264, opened 2019-06-19
DIT-1053 anchor tags with button role

SteveSugden:


SteveSugden:
Copied that loop from govuk-template.js .  But I agree, the variable name is not good

PR #269, opened 2019-06-20
Permissions

ed0906:


ed0906:
There's still some left, I just dont want this to become a mega PR

ed0906:
> Browser won't let me comment on individual change for some reason...
> 
> In controllers, e.g SuspendCaseController -> isValidCase override just returns true for quite a few of them, can you add a default implementation to the parent class, and remove the overrides?

Good idea, will do

PR #272, opened 2019-06-20
DIT-1120 - Page titles for queues

aaroncarroll82:


aaroncarroll82:
> Happy if it works. Does it break any ATs?

I think it probably will, I'll run them and update the ones we need

PR #291, opened 2020-01-07
DIT-1166 updated commodity codes

colesadam:


colesadam:
pulling this as csv format is incorrect - Esther to review

PR #299, opened 2020-02-19
DIT-1528 | Updated not_authorised page content and moved text to messages

periklas:


colesadam:
@periklas @shahzadbhatti1 I'm confused. My reading of the "givenAuthFailed()" function in the tests is that the status returned should be "UNAUTHORIZED", not "OK". Secondly we should only be testing the content of the unauthorised page once.

PR #304, opened 2020-03-19
DIT-1218 Added toggle for LiabilityController v2 

periklas:
To run: ```sbt -Dtoggle.new-liability-details=true run```

### DIFF
--- a/app/models/viewmodels/C592ViewModel.scala
+++ b/app/models/viewmodels/C592ViewModel.scala
@@ -1,4 +1,4 @@
-@*
+/*
  * Copyright 2020 HM Revenue & Customs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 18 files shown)


app/models/viewmodels/C592ViewModel.scala
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.viewmodels
+
+import models.Case
+
+case class C592ViewModel(entryNumber: String)
```
abhijeetardale: can we change view name to more meaningful?

test/unit/controllers/v2/LiabilityControllerSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+      when(inject[liability_view].apply(any())(any(), any(), any())) thenReturn Html("body")
+
+      val result: Future[Result] = route(app, FakeRequest("GET", "/manage-tariff-classifications/cases/v2/123456/liability")).get
+
+      status(result) shouldBe OK
+
+      val expected = LiabilityViewModel.fromCase(Cases.liabilityCaseExample)
+      verify(inject[liability_view], times(1)).apply(meq(expected))(any(), any(), any())
+
+      println(contentAsString(result))
```
abhijeetardale: please remove println

app/models/viewmodels/C592ViewModel.scala
colesadam: Not sure what, it models the C592 view :) That's how it's displayed to the end user

PR #311, opened 2020-03-25
DIT-1819 | Action case button

gorkemunsal:


test/util/utils/Cases.scala
```diff
@@ -57,6 +57,7 @@ object Cases {
   )
   val btiCaseWithExpiredRuling = btiCaseExample.copy(status = CaseStatus.COMPLETED, decision = Some(expiredRuling))
   val liabilityCaseWithExpiredRuling = liabilityCaseExample.copy(status = CaseStatus.COMPLETED, decision = Some(expiredRuling))
+  val operatorWithoutPermissions = Operator(id = "0", name =Some("liability op name"), permissions = Set())
```
colesadam: Does this really belong in the Cases util? Can you not just put it in the spec file?

gorkemunsal: It's in the testing package. This is the file where we have all the case examples that are being used in the spec files. We thought this might be used in other spec files as well that's why we created one but happy to move it to the spec file?

colesadam: OK - I think it's because it's in a class called Cases - seems like anything that isn't a Case shouldn't be accessible outside, only accessed internally where needed to construct a "Case". However I think it's already a bit polluted.

PR #314, opened 2020-04-02
DIT-1930 Added ActivityTab

periklas:


app/controllers/v2/LiabilityController.scala
```diff
@@ -65,4 +74,39 @@ class LiabilityController @Inject()(
       Some(AttachmentsTabViewModel(liabilityCase.reference, applicantFiles, letter, nonApplicantFiles))
     }
   }
+
+
+  def addNote(reference: String): Action[AnyContent] = (verify.authenticated andThen
+    verify.casePermissions(reference) andThen verify.mustHave(Permission.ADD_NOTE)).async { implicit request =>
+    def onError: Form[ActivityFormData] => Future[Result] = errorForm => {
+      val liabilityCase: Case = request.`case`
```
colesadam, line 82: much of this code is common with the main get function - can you factor it out?

gorkemunsal: we've moved it to a separate function

colesadam: surely this is all common as well?

` val liabilityCase: Case = request.`case`
      val liabilityViewModel = LiabilityViewModel.fromCase(liabilityCase, request.operator)
      for {
        tuple <- liabilityViewActivityDetails(reference)
        attachmentsTab <- getAttachmentTab(liabilityCase)
        c592 = Some(C592ViewModel.fromCase(liabilityCase))
        activityTab = ActivityViewModel.fromCase(liabilityCase, tuple._1, tuple._2)
      } yield {
        Ok(liability_view(liabilityViewModel, c592, attachmentsTab, activityTab, activityForm))
      }`

The only difference is whether the form is empty or contains errors?

PR #315, opened 2020-04-06
Dit 1931

frikit:


app/views/v2/liability_view.scala.html
```diff
@@ -70,11 +76,13 @@
             }
 
             @* activity*@
+            @if(activityViewModel.isDefined) {
```
periklas, line 79: Does that need to be an Option, since this tab is not optional?

frikit: just for test to be easier to test view itself

PR #320, opened 2020-04-09
Dit 1817

frikit:


app/models/viewmodels/LiabilityViewModel.scala
```diff
@@ -39,10 +40,17 @@ object LiabilityViewModel {
...
+      operator.permissions.contains(Permission.COMPLETE_CASE)
+
+    def changeCaseStatus: Boolean =
+      c.status == CaseStatus.OPEN && completeCasePermission
+
     /**
-     * open/referred/rejected/suspended/completed
-     * @return true || false
-     */
+      * open/referred/rejected/suspended/completed
```
abhijeetardale, line 50: can we remove commented code?

frikit: removed

PR #321, opened 2020-04-09
DIT-1951 Update tabs CSS & JS

adamliptrot-oc:


adamliptrot-oc:
@colesadam Nope, it should never have been put in in the first place.

PR #324, opened 2020-04-15
DIT-1950 Added Keywords tab

periklas:


test/unit/models/viewmodels/LiabilityViewModelSpec.scala
```diff
@@ -67,25 +63,38 @@ class LiabilityViewModelSpec extends UnitSpec {
 
   "showActionThisCase" should {
 
-    "not show action this case button when isNewCase = false and hasPermissions = false" in {
```
frikit, line 70: Duplicate test after merge with master

test/unit/views/v2/LiabilityViewSpec.scala
```diff
@@ -202,9 +226,126 @@ class LiabilityViewSpec extends ViewSpec {
...
+        activityForm,
+        None,
+        uploadAttachmentForm,
+        emptyKeywordsTabViewModel,
+        keywordForm)
+      )
+      doc shouldNot containElementWithID("keywords_tab")
+    }
+
+    "render the action this case button for new case" in {
```
frikit, line 269: all test for buttons was moved to new partial view and with separate test cases here https://github.com/hmrc/tariff-classification-frontend/blob/master/test/unit/views/partials/liabilities/CaseActionButtonsViewSpec.scala make sure you cleanup here as well

colesadam: removed button tests from liabilityviewspec

PR #333, opened 2020-04-30
DIT-1813 Added Edit-C592 form with extra fields

frikit:


app/models/forms/LiabilityDetailsForm.scala
```diff
@@ -45,20 +45,22 @@ object LiabilityDetailsForm {
 
   private def validEmailFormat(email: String): Boolean = email.trim.isEmpty || emailRegex.findFirstMatchIn(email.trim).nonEmpty
 
-  private def form2Liability(existingLiability: LiabilityOrder): (Option[Instant], String, Option[String], Option[String], Option[String], Option[String], String, String, Option[String]) => LiabilityOrder = {
+  private def form2Liability(existingCase: Case): (Option[Instant], String, Option[String], Option[String], Option[String], Option[String], String, String, Option[String]) => Case = {
     case (entryDate, traderName, goodName, entryNumber, traderCommodityCode, officerCommodityCode, contactName, contactEmail, contactPhone) =>
-      existingLiability.copy(
+      existingCase.copy(application = existingCase.application.asLiabilityOrder.copy(
```
DavidGregory084, line 50: FWIW we could try using https://github.com/softwaremill/quicklens to get rid of nested `.copy` stuff

app/models/forms/mappings/Constraints.scala
```diff
@@ -18,8 +18,12 @@ package models.forms.mappings
 
 import play.api.data.validation.{Constraint, Invalid, Valid}
 
+import scala.util.matching.Regex
+
 trait Constraints {
 
+  private val emailRegex = """^[a-zA-Z0-9\.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$""".r
```
DavidGregory084, line 25: Is there maybe an internal library for doing this already?

app/models/forms/v2/LiabilityDetailsForm.scala
```diff
@@ -0,0 +1,181 @@
...
+import models.forms.mappings.Constraints
+import models.forms.mappings.FormMappings._
+import models._
+import play.api.data.Form
+import play.api.data.Forms._
+
+object LiabilityDetailsForm extends Constraints {
+
+  private val numbersOnlyRegex = """^\d+$""".r
+  private val btiRefRegex = """[0-9]{6,22}""".r
```
DavidGregory084, line 32: Maybe we should put validation regexes for things relating to ATaR (had to look that up) data in a utility class?

DavidGregory084: e.g. similar to `FormDate`

app/models/forms/LiabilityDetailsForm.scala
frikit: for me copy is more readable, I don't think we need to add new lib specifically for this stuff

maybe in future

thanks

app/models/forms/mappings/Constraints.scala
frikit: there is a internal library but is not very robust and in case if business need specific errors in specific cases you can't do this in library, this is why we kept this regex in place here

app/models/forms/v2/LiabilityDetailsForm.scala
frikit: extract these lines into methods for each case in form constraints

frikit:
Fix PR comments

PR #335, opened 2020-05-01
DIT-1814 Fix bug with display of edit rulings link

DavidGregory084:
This issue arose because the logic checking whether to display the link was duplicated between the view model and the permission, so we have moved all the logic into the permission.

### DIFF
--- a/test/unit/models/viewmodels/RulingViewModelTest.scala
+++ b/test/unit/models/viewmodels/RulingViewModelTest.scala
@@ -27,28 +27,10 @@ class RulingViewModelTest extends UnitSpec {
 
   "showEditRuling" should {
 
-    "show edit ruling button when case status is OPEN" in {
+    "show edit ruling button when we have EDIT_RULING permission" in {
       val aCase = dummyCase.copy(status = OPEN)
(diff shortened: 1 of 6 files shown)


test/unit/models/viewmodels/RulingViewModelTest.scala
```diff
@@ -27,30 +27,12 @@ class RulingViewModelTest extends UnitSpec {
...
-      val aCase = dummyCase.copy(status = SUSPENDED)
-      val permissions: Set[Permission] = Set(Permission.EDIT_RULING)
-      RulingViewModel.fromCase(aCase, permissions).showEditRuling shouldBe true
-    }
-
-    "not show edit ruling button because case status" in {
-      val aCase = dummyCase.copy(status = REJECTED)
-      val permissions: Set[Permission] = Set(Permission.EDIT_RULING)
-      RulingViewModel.fromCase(aCase, permissions).showEditRuling shouldBe false
-    }
```
DavidGregory084, line 52: These tests are irrelevant now because it's only the permission which determines whether to display the link

PR #340, opened 2020-05-06
DIT-1857: Fix appeal status in case header

frikit:


app/models/Appeal.scala
```diff
@@ -19,9 +19,21 @@ package models
...
+object Appeal {
+
+  def highestAppealFromDecision(decision: Option[Decision]): Option[Appeal] = {
+    val appeals = decision.map(_.appeal).getOrElse(Seq.empty)
+
+    if (appeals.nonEmpty) {
+      Some(appeals.sortWith((a, b) => a.`type`.id > b.`type`.id).head)
+    } else {
+      None
+    }
```
DavidGregory084, line 37: I think this could be implemented with `appeals.maxBy(_.``type``.id)`

`maxByOption` would allow you to do this even more easily, but unfortunately that was only introduced in Scala 2.13

For reference:
```scala
for {
  decision <- decisionOpt
  highestAppeal <- decision.appeals.maxByOption(_.`type`.id)
} yield highestAppeal
```

app/models/AppealType.scala
```diff
@@ -32,6 +32,9 @@ object AppealType extends Enumeration {
     case _ => "Appeal"
   }
 
+  /**
+   * The order of enum matters as it is used how to show elements in UI in some cases, where it is sorted by ID
+   */
   type AppealType = Value
   val ADR, REVIEW, APPEAL_TIER_1, APPEAL_TIER_2, COURT_OF_APPEALS, SUPREME_COURT = Value
```
DavidGregory084, line 39: Just an idea, but could we avoid making enumeration order significant by using a sealed abstract class or something like that?

```
sealed abstract class AppealType(appealTier: Int) extends Product with Serializable

case object ADR extends AppealType(0)

// etc.
```

test/unit/models/AppealTest.scala
```diff
@@ -0,0 +1,114 @@
...
+      val c = aCase(
+        withoutDecision()
+      )
+
+      // When
+      val res = Appeal.highestAppealFromDecision(c.decision)
+
+      // Then
+      res shouldBe None
+    }
```
DavidGregory084, line 41: ScalaTest has a trait [GivenWhenThen](http://doc.scalatest.org/3.0.1-2.12/org/scalatest/GivenWhenThen.html) which you can use if you want BDD style output (see also [FeatureSpec](http://doc.scalatest.org/3.0.1-2.12/org/scalatest/FeatureSpec.html))

app/models/AppealType.scala
frikit: I tried to change the enums in project to be strictly defined but from/to json start to have a lot of issues this is why I just added comment for now

test/unit/models/AppealTest.scala
frikit: It was just copy from other class :),  I will keep for now as is

app/models/Appeal.scala
frikit: great comment I will implement this, thanks

app/models/AppealType.scala
DavidGregory084: Ah yeah I forgot that play-json doesn't support sealed trait / sealed abstract class

PR #348, opened 2020-05-28
DIT-1818: Add search cases button in liabilities case header

frikit:


app/models/viewmodels/LiabilityViewModel.scala
```diff
@@ -16,29 +16,40 @@
...
-    isNewCase && hasPermissions
-  }
+  def isNewCase: Boolean = caseStatus == CaseStatus.NEW
+
+  def showRulingAndKeywordsTabs: Boolean =
+    caseStatus == CaseStatus.OPEN ||
+      caseStatus == CaseStatus.REFERRED ||
+      caseStatus == CaseStatus.REJECTED ||
+      caseStatus == CaseStatus.SUSPENDED ||
+      caseStatus == CaseStatus.COMPLETED
```
DavidGregory084, line 38: Maybe we could use `Set(CaseStatus.OPEN, CaseStatus.REFERRED, /* etc */).contains(caseStatus)` here?

frikit: If you think it does make big difference I can make this change, to me there is no difference

frikit: done :)

DavidGregory084: Haha I'm not sure it makes a big difference either. :) I have a distant memory that we already have Sets of statuses somewhere in the code.

PR #360, opened 2020-07-07
DIT 1895 fixed issues on Select files to attach to the ruling

shiromi:


app/views/partials/attachments_edit_list.scala.html
```diff
@@ -0,0 +1,69 @@
...
+@import models.Case
+@(id: String, attachments: Seq[StoredAttachment], showSelect: Boolean = false,
+showRemoval: Boolean = false, caseRef: Option[String] = None, startAtTabIndex : Option[Int] = None)(implicit request: AuthenticatedRequest[_], messages: Messages)
+
+@if(attachments.nonEmpty){
+<legend class="w-100">
+<div id="@{id}-div">
+    <div class="border-bottom display-flex row-alignment" >
+        <div class="bold-small w-75 pl-0">File</div>
+        @if(showSelect) {<div class="bold-small w-25 p-0" id="@{id}-header-select" >Publish</div>}
```
frikit, line 32: based on file name and the logic which is similar in the attachments_list... maybe we dont need showselect param? because it will be always true for this file? I assume... IF not it is ok

test/unit/views/partials/AttachmentsEditListViewSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+        url = Some("url"),
+        scanStatus = Some(ScanStatus.READY),
+        timestamp = ZonedDateTime.of(2019,1,1,0,0,0,0,ZoneOffset.UTC).toInstant
+      )
+
+      // When
+      val doc = view(attachments_edit_list("MODULE", Seq(attachment)))
+
+      // Then
+      doc should containElementWithID("MODULE-div")
```
frikit, line 54: maybe to check for checkbox on page as well?

app/views/partials/attachments_edit_list.scala.html
shiromi: Thanks Victor. I have removed showselect param as it is no need anymore.

app/views/partials/attachments_edit_list.scala.html
```diff
@@ -29,7 +29,7 @@
 <div id="@{id}-div">
     <div class="border-bottom display-flex row-alignment" >
         <div class="bold-small w-75 pl-0">File</div>
-        @if(showSelect) {<div class="bold-small w-25 p-0" id="@{id}-header-select" >Publish</div>}
```
frikit: You just did remove the if statement not parameter itself

frikit:
Now looks good to me

PR #362, opened 2020-07-14
DIT-1933 added indications that the pdf will open in a new window on …

shiromi:
…the Item and Ruling tabs of a BTI case


app/views/partials/submitted_application_link.scala.html
```diff
@@ -21,5 +21,5 @@
 
 <div id="submitted-application" class="mb-3">
    If this case is under review or appeal,
-   <a  @if(useTabIndex.isDefined){ tabindex="@{useTabIndex.get}" } id="submitted-application-link" target="_blank" href="@PdfDownloadController.applicationPdf(c.reference)">  view and print a PDF of the original application</a>.
+   <a  @if(useTabIndex.isDefined){ tabindex="@{useTabIndex.get}" } id="submitted-application-link" target="_blank" href="@PdfDownloadController.applicationPdf(c.reference)">  view and print a PDF of the original application (opens in a new window or tab).</a>.
```
vajarova: is it better to keep using messages for consistency? @DavidGregory084

DavidGregory084: Yeah that's probably a better solution 👍

PR #369, opened 2020-07-29
DIT 1829 url correction

shiromi:


conf/application.conf
```diff
@@ -187,7 +187,7 @@ mongodb {
 accessibility-urls {
   accessibility = "binding-tariff-application/accessibility"
   govukAccessibilityUrl = "https://www.gov.uk/help/accessibility"
-  subdomainUrl = "https://www.gov.uk/tax.service.gov.uk/manage-tariff-classifications"
+  subdomainUrl = "/tax.service.gov.uk/manage-tariff-classifications"
```
vajarova, line 190: we still need the https://www here 
so it will be https://www.tax.service.gov.uk/manage-tariff-classifications

PR #376, opened 2020-08-17
DIT-2166 fixed mobile view issue on file attachment section

shiromi:


adamliptrot-oc:
As this is using Assets Frontend, which consumes Elements and in turn the GOVUK Frontend Toolkit, you could look at using the media query helpers. It depends on how the assets are bundled/imported as to whether you can access them from the service's own SASS files. https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_conditionals.scss#L12

A couple of questions:
Where are the values for the padding coming from?
If this padding is ok on mobile, why is it not ok on desktop?

colesadam:
Does this need any further thought based on Adam's comments above? @shiromi

DavidGregory084:
It doesn't seem like anyone is going to merge this

PR #379, opened 2020-08-25
DIT-2270 | Create case event added

gorkemunsal:
This repo contains the create case event for liabilities.

This needs to be merged before this PR: https://github.com/hmrc/binding-tariff-classification/pull/139


app/service/CasesService.scala
```diff
@@ -416,6 +420,15 @@ class CasesService @Inject()(appConfig: AppConfig,
     addEvent(original, updated, details, operator)
   }
 
+  private def addCaseCreatedEvent(liabilityCase: Case, operator: Operator)
+                                 (implicit hc: HeaderCarrier): Future[Unit] = {
+    val event = NewEventRequest(CaseCreated("Liability case created"), operator)
+    connector.createEvent(liabilityCase, event) recover {
+      case t: Throwable =>
+        Logger.error(s"Could not create Event for case [${liabilityCase.reference}] with payload [$event]", t)
+    } map (_ => ())
```
DavidGregory084, line 429: I think we could reuse the existing `addEvent` here
```suggestion
    val details = NewEventRequest(CaseCreated("Liability case created"), operator)
    addEvent(liabilityCase, liabilityCase, details, operator)
```

app/views/partials/activity_details_events_table.scala.html
```diff
@@ -132,6 +132,14 @@
                                 useTabIndex = Some(startAtTabIndex + (index*2))
                             )
                         }
+
+                        case EventType.CASE_CREATED => {
+                            val body = event.details.asInstanceOf[CaseCreated]
+                                <div id="activity-events-row-case_created">
+                                <span id="activity-events-row-application_submitted-comment" class="pre-wrap">{body.comment}</span>
+                                </div>
```
DavidGregory084, line 140: Should we use the `progressive_disclosure` partial here in order to preserve the tabindex functionality? Or even define a new partial `activity_case_created` similar to `activity_case_completed`?

app/views/partials/liabilities/activity_tab_events_table.scala.html
```diff
@@ -60,6 +60,15 @@ <h3 class="heading-medium mt-30 mb-30">@messages("case.v2.liability.activity_tab
                         case EventType.CASE_COMPLETED => {
                             activity_case_completed(index, event)
                         }
+
+                        case EventType.CASE_CREATED => {
+                        val body = event.details.asInstanceOf[CaseCreated]
+
+                        <div id="activity-events-row-case_created">
+                            <span id="activity-events-row-application_submitted-comment" class="pre-wrap">{body.comment}</span>
+                        </div>
```
DavidGregory084, line 69: Same question here

DavidGregory084 (approved):

app/service/CasesService.scala
gorkemunsal: We didn't use that one because that takes the original and the updated case but on our scenario we are only just creating the case so have only 1 case object hence we created the addCaseCreatedEvent that takes only the case. However I suppose we can use the addEvent method and pass the case as a original and updated?

app/views/partials/activity_details_events_table.scala.html
periklas: Good point, it now uses progressive_disclosure 👍

app/views/partials/liabilities/activity_tab_events_table.scala.html
periklas: Good point, it now uses progressive_disclosure 👍

PR #381, opened 2020-08-27
DIT-2264 | Removed importOrExport parameter from model, view and tests

periklas:


periklas:
We should update the journey tests before this can be merged.

shiromi (approved):
Looks good to me

PR #382, opened 2020-08-27
DIT-2288 removed extra space to on aria-describedby id to correct the…

shiromi:
… element id


OC-Pomaroff (approved):
Looks fine for now, although it might be worth breaking this logic into a function block to make it a bit easier to read as we're checking the same conditions multiple times in this long statement.

PR #385, opened 2020-09-04
DIT 2289 fixed some issues Operational service from an accessibility point of view

shiromi:
Edit details currently H1 and it should be H2
Upload new files currently H2 and it should be H3
Add 'LIVE' status for user
tabindex doesn't wok on BTI complete cases


DavidGregory084 (approved):
Looks good to me! 👍

DavidGregory084:
It looks like this may require some changes to the acceptance tests @shiromi so I don't think we can merge right away

OC-Pomaroff (approved):

PR #396, opened 2020-10-19
DIT-2331 | Modified to cater for List of relatedBTIReferences as well…

warrenskiwork1979:
… as singular


DavidGregory084 (approved):

DavidGregory084:
@warrenskiwork1979 I think it is probably worth changing the JSON formatters so that they read default values if either of these properties isn't populated while we transition from one to the other

PR #397, opened 2020-10-21
DIT-2319 Classification view front end work

shiromi:
This ticket is only Frond end implementation. I have added small unit test script to avoid fail the unit test, but we need to add the proper unit test while we do the backend implementation.


DavidGregory084 (approved):
Looks really great to me 👍

vajarova (approved):

PR #403, opened 2020-11-16
DIT-2440 | new view for my/referred/gateway cases on dashboard

gorkemunsal:


app/views/components/common_atar.scala.html
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import models.request.AuthenticatedRequest
+@import config.AppConfig
+
+@()(implicit request: AuthenticatedRequest[_], messages: Messages, appConfig: AppConfig)
+
+ <table id="atars">
```
OC-Pomaroff, line 22: The table in atars, liabilities, correspondence, and misc are all the same.
The only difference is the cases that we will pass through, and the ID's I suppose.

Could we make it a single component where we pass in an id-prefix, and later the cases?

app/views/v2/common_cases_view.scala.html
```diff
@@ -0,0 +1,95 @@
...
+@main(bodyClasses = None, customTitle = Some(messages(title))) {
+
+    @primary_navigation()
+    @my_cases_secondary_nav()
+
+    <!--the heading component-->
+    @heading("heading text")
+
+    <div id="common-cases-tabs" class="govuk-tabs" data-module="govuk-tabs">
+        <ul class="govuk-tabs__list">
```
OC-Pomaroff, line 40: Is there a way to make this tabbed-view a more generic component so that it can  be re-used with different lists of tabs in all of the dashboard views?

It's likely that the list of case types won't change, but in the future queues will be able to be added/removed/renamed so this will need to be dynamically driven somehow.

OC-Pomaroff (commented):
Can we make some of these components more generic?

app/views/components/common_atar.scala.html
periklas: It seems from the prototype that the tables have different headers, hence the need for different components.

OC-Pomaroff: Ah my bad, you are correct!

app/views/v2/common_cases_view.scala.html
OC-Pomaroff: As discussed with @periklas we can look at iterative improvements/generalisations as we do the subsequent tickets!

OC-Pomaroff (approved):

PR #404, opened 2020-11-19
Dit 2441 - views for secondary navigation for star, liability , corresp

emilia-trendafilova:


app/views/components/primary_navigation.scala.html
```diff
@@ -14,22 +14,47 @@
...
         <div class="moj-primary-navigation__nav">
             <nav class="moj-primary-navigation" aria-label="Primary navigation">
                 <ul class="moj-primary-navigation__list">
                     <li class="moj-primary-navigation__item">
-                        <a class="moj-primary-navigation__link" aria-current="page"  href="#">Home</a>
+                        @if(selectedTab == HomeTab) {
+                        <a id ="primary_nav_home_tab" class="moj-primary-navigation__link"  aria-current="page" href="#">Home</a>
+                        } else {
+                        <a id ="primary_nav_home_tab" class="moj-primary-navigation__link"  href="#">Home</a>
+                        }
```
DavidGregory084, line 36: No need to do this now, but it would be good to have a helper function to add aria-current attribute similar to [this](https://github.com/playframework/playframework/blob/2.8.x/core/play/src/main/scala/views/html/helper/CSPNonce.scala#L46). 

It would also be good to componentise these elements to avoid the repetition of all of these CSS classes.

DavidGregory084 (approved):
Looks good 👍

PR #411, opened 2020-11-27
Dit 2444 - Added queries for all open ATaR cases and tests

warrenskiwork1979:
Must merge this first: https://github.com/hmrc/binding-tariff-classification/pull/153


app/connector/BindingTariffClassificationConnector.scala
```diff
@@ -73,6 +73,20 @@ class BindingTariffClassificationConnector @Inject() (
     s"${appConfig.bindingTariffClassificationUrl}/cases?$queryString"
   }
 
+  private def buildQueryAllQueuesURL(
```
OC-Pomaroff, line 76: Is this function necessary?

If we change `buildQueryUrl` to accept a sequence of queueId's then it's the same function, since `queueIds.mkString(",")` for a single queueId will just return the queueId itself.

OC-Pomaroff (approved):

OC-Pomaroff (approved):

OC-Pomaroff (approved):
LGTM

OC-Pomaroff:
PS. it'd be nice to set your editor to use scalafmt since the rest of the project is formatted

app/connector/BindingTariffClassificationConnector.scala
oliwiaKot: you're right, thanks! We did as you suggested, removed the unnecessary method and just kept one that accepts a sequence of queueIds

PR #423, opened 2020-12-09
DIT- 2461 ATaR headers & navigation

emilia-trendafilova:


app/views/components/primary_navigation.scala.html
```diff
@@ -60,15 +60,13 @@
             </nav>
         </div>
         <div class="moj-primary-navigation__search">
-            <div class="moj-search moj-search--ondark moj-search--inline"><form action="/205-classification/search/advanced-search-results" method="get">
+            <div class="moj-search moj-search--ondark moj-search--inline">
+                @helper.form(action = controllers.routes.SearchController.search(), 'novalidate -> "novalidate") {
```
OC-Pomaroff, line 64: Should we be using uk.gov.hmrc.play.views.html.helpers.form?
I can't recall if play's helper.form includes the CSRF token or if that was version dependent.

app/views/partials/case_heading.scala.html
```diff
@@ -22,12 +22,12 @@
 @(c: Case, displayBackLink: Boolean = true)(implicit messages: Messages, request: Request[_])
 
 <div id="case-heading" class="case-heading mt-1 mb-2">
-
+    <br>
```
OC-Pomaroff, line 25: Could we just increase the bottom margin instead of adding a line break?

app/views/v2/case_heading.scala.html
```diff
@@ -34,6 +34,7 @@
 }
 
 <div id="case-heading" class="case-heading mt-1 mb-2">
+  <br>
```
OC-Pomaroff, line 37: Same here

OC-Pomaroff (approved):

app/views/components/primary_navigation.scala.html
periklas: It includes the CSRF token. We have replaced it with the uk.gov.hmrc.play.views.html.helpers.form 👍

app/views/partials/case_heading.scala.html
periklas: Of course, we have changed it! Thanks! 👍

DavidGregory084 (approved):
This looks great to me 👍 the only thing I wonder is, do we need `NoTabSelected` or could `HomeTab` serve the same purpose of being the default tab?

periklas:
> This looks great to me 👍 the only thing I wonder is, do we need `NoTabSelected` or could `HomeTab` serve the same purpose of being the default tab?

That was the initial approach (using the HomeTab as a default tab). Although there are pages that are not part of the navigation (i.e. Accessibility page) which probably should not be under any tab selected.

PR #433, opened 2020-12-17
Dit 2446 - Queries for open correspondence cases tile

emilia-trendafilova:


app/models/viewmodels/CasesTabViewModel.scala
```diff
@@ -135,14 +110,12 @@ object CasesTabViewModel {
...
   )
 
-
   //TODO remove dummy correspondence example when writing queries for the tabs
   val corrApplicationExample: CorrespondenceApplication = CorrespondenceApplication(
-   Some("Starter"),
+    Some("Starter"),
     Some("Agent 007"),
     Address("New building", "Old Town", None, None),
     Contact("a name", "anemail@some.com", None),
```
DavidGregory084, line 121: Could we remove these examples as well now?

DavidGregory084 (approved):
Looks great 💯

vajarova (approved):

app/models/viewmodels/CasesTabViewModel.scala
periklas: 👍

kuldeepwadhwa9 (approved):

PR #434, opened 2020-12-20
DIT-2522 Convert ATaR case view to use new tab component

DavidGregory084:
Convert ATaR case view to use the same tab component as the liability case view and to use the same approach for building the view. There is still a huge amount of duplication and multiple different approaches to solving the same problem all over the place but at least the case view pages are built the same.


DavidGregory084:
I think the build is failing because of changes on the base branch (DIT-2525) - this should be able to be re-run once those failures are fixed

PR #449, opened 2021-01-05
Dit 2536 | Correspondence change case status, refer and completed

warrenskiwork1979:


app/views/v2/case_heading.scala.html
```diff
@@ -44,5 +44,16 @@ <h1 class="heading-xlarge case-title">
         }
     </h1>
 
+    @if(caseHeader.caseType == ApplicationType.CORRESPONDENCE) {
+        <div id="contact-section" class="mb-1">
+            <div class="bold">
+                Case source - @caseHeader.businessName
```
OC-Pomaroff, line 50: Should this be using `caseHeader.caseSource` like the other case_heading view?

app/views/refer_reason_checkbox.scala.html
```diff
@@ -0,0 +1,46 @@
...
+
+@(form: Form[CaseReferral])(implicit messages: Messages)
+
+@isChecked(f1: Option[String], f2: Option[String], matchText: String) = @{
+    matchText == f1.getOrElse("") || matchText == f2.getOrElse("")
+}
+
+@input_checkbox_group(
+    field = form("reasons"),
+    legend = Some("Why are you referring this case?"),
```
OC-Pomaroff, line 28: Should we move these strings into the messages file?

app/models/CaseStatusRadioInput.scala
```diff
@@ -16,26 +16,33 @@
...
+  case object Reject extends CaseStatusRadioInput("reject", Set(ATAR, LIABILITY))
+  case object Suspend extends CaseStatusRadioInput("suspend", Set(ATAR, LIABILITY))
+  case object MoveBackToQueue extends CaseStatusRadioInput("move_back_to_queue", ApplicationType.values)
+  case object Release extends CaseStatusRadioInput("release", ApplicationType.values)
+  case object Suppress extends CaseStatusRadioInput("suppress", ApplicationType.values)
+
+  private val changeCaseStatusValues: Seq[CaseStatusRadioInput]  = Seq(Complete, Refer, Reject, Suspend, MoveBackToQueue)
+  private val releaseOrSuppressValues: Seq[CaseStatusRadioInput] = Seq(Release, Suppress)
+
+  def changeCaseStatusOptionsFor(applicationType: ApplicationType): Seq[InputRadio] =
```
OC-Pomaroff, line 42: It seems that this code is unused - was the intention to use it in the refer_case view?

OC-Pomaroff (approved):

app/models/CaseStatusRadioInput.scala
warrenskiwork1979: @OC-Pomaroff This code is used in the change case status view

OC-Pomaroff: Whoops so it is, good point!

kuldeepwadhwa9 (approved):

PR #451, opened 2021-01-07
Dit 2539 - View correspondence case

emilia-trendafilova:


app/models/viewmodels/ApplicationsTab.scala
```diff
@@ -53,16 +53,21 @@ object ApplicationsTab {
 
   def assignedToMeCases(cases: Seq[Case]): ApplicationTabViewModel = {
 
-    val atars = cases.filter(aCase => aCase.application.isBTI && aCase.status == CaseStatus.OPEN)
+    val assignedCases =
+      cases.filter(aCase => aCase.status == CaseStatus.OPEN)
```
OC-Pomaroff, line 57: Nice, maybe this initial filter of caseType could be repeated below for gateway cases, too?

app/views/components/correspondenceCaseTable.scala.html
```diff
@@ -47,7 +47,12 @@
     @defining(result.application.asCorrespondence) { correspondence =>
 
      <tr id="@{caseTab.tabMessageKey}-details-@index">
-      <td id="@{caseTab.tabMessageKey}-reference-@index"><a href="@routes.CaseController.get(result.reference)">@{result.reference}</a></td>
+      <td id="@{caseTab.tabMessageKey}-reference-@index">
+      @if(result.queueId.isDefined && !result.assignee.map(_.id).contains(request.operator.id)) {
+       <a href="@routes.AssignCaseController.get(result.reference)">@result.reference</a>
```
OC-Pomaroff, line 52: Does this need to be `@{result.reference}` (with the curly braces?)

app/views/forms/correspondence_sending_sample_form.scala.html
```diff
@@ -0,0 +1,44 @@
...
+        id = "sample_status_form",
+        preSelect = false,
+        label = Some(messages("case.liability.sending.sample.change-return-header")),
+        labelClass = Some ("heading-large")
+    )
+    @confirm_or_cancel_buttons(
+        id = "sending_sample_form-submit",
+        confirmText = messages("case.liability.sending.sample.change-return-confirm"),
+        cancelText = messages("case.liability.sending.sample.change-return-cancel"),
+        cancelHref = routes.CaseController.sampleDetails(c.reference).withFragment(Tab.SAMPLE_TAB.name)
```
OC-Pomaroff, line 41: Just to check, are these messages intended to shared with `case.liability.___` ?

app/views/partials/sample/sample_details_correspondence.scala.html
```diff
@@ -0,0 +1,89 @@
...
+)
+
+@(s: SampleStatusTabViewModel, startAtTabIndex : Int = 0)(implicit request: AuthenticatedRequest[_], messages: Messages, appConfig: AppConfig)
+
+@tab_heading("samples", messages("case.v2.liability.sample_status_tab.tab_title"), Some(messages("case.v2.liability.sample_status_tab.hint")))
+
+<dl class="dl-mb-2">
+
+
+    <div id="liability-sending-samples" class="grid-row">
```
OC-Pomaroff, line 36: A few id's are still using liability here

OC-Pomaroff (approved):
LGTM!

app/views/forms/correspondence_sending_sample_form.scala.html
emilia-trendafilova: These messages are exactly the same as liabilities so to avoid duplication we have reused them.

OC-Pomaroff: Sounds reasonable!

app/views/partials/sample/sample_details_correspondence.scala.html
emilia-trendafilova: Thanks good spot 👍

app/controllers/SampleController.scala
```diff
@@ -51,10 +51,13 @@ class SampleController @Inject() (
...
   ): Html =
-    if (options.contains("liability"))
+    if (options.contains("liability")) {
       views.html.change_liablity_sending_sample(c, notFilledForm)
-    else
+    } else if (options.contains("correspondence")) {
+      views.html.change_correspondence_sending_sample(c, notFilledForm)
+    } else {
       views.html.change_sample_status(c, notFilledForm)
+    }
```
DavidGregory084, line 60: Not necessarily something we necessarily need to change in this PR, but do you think we need this concept of `options` at all? It seems like we could just match on the `` c.application.`type` `` here

emilia-trendafilova: Good spot!. We have changed it now to use application.'type' instead of options.

PR #454, opened 2021-01-11
DIT-2619 Update parameters to case completed email

DavidGregory084:


app/service/EmailService.scala
```diff
@@ -19,23 +19,30 @@ package service
...
     }
 
     val email: CaseCompletedEmail = CaseCompletedEmail(
       Seq(c.application.contact.email),
-      CaseCompletedEmailParameters(c.application.contact.name, c.reference, c.application.asATAR.goodName)
+      CaseCompletedEmailParameters(
+        recipientName_line1 = c.application.contact.name,
+        reference = c.reference,
+        goodsName = c.application.asATAR.goodName,
+        officerName = operator.name.getOrElse(""),
```
DavidGregory084, line 43: There is no good alternative that we can use here. It will render as an empty line in the email, i.e. something like `Yours sincerely,   Tariff Classification Team`

oliwiaKot (approved):

OC-Pomaroff (approved):

PR #460, opened 2021-01-14
Activity tab header change from "name" to "identifier"

oliwiaKot:
changed the header from name to identifier and also display operator id when its present rather than always displaying name as per prototype


DavidGregory084 (approved):

DavidGregory084:
Do we really want to display the ID number in preference to the name @oliwiaKot?

PR #462, opened 2021-01-14
Dit 2553 | Modified Miscellaneous case refer confirm screen and code  reuse from work on Correspondence refer

warrenskiwork1979:


DavidGregory084 (approved):
Looks great to me 👍

PR #465, opened 2021-01-19
DIT-1959 | adding extra fields and mappings to edit c592 page

gorkemunsal:


app/views/partials/error_summary.scala.html
```diff
@@ -22,7 +22,7 @@ <h2 class="heading-medium error-summary-heading" id="error-summary-heading">
         </h2>
         <ul role="list" class="error-summary-list">
             @for((error: FormError, index: Int) <- errors.zipWithIndex) {
-                <li><a href="#@{error.key}" tabindex="@startAtTabIndex.map(x => x+2+index)">@messages(error.message, error.args:_*)</a></li>
+                <li><a href="#@error.key.replaceAll("""[\.\[\]]""", "_")" tabindex="@startAtTabIndex.map(x => x+2+index)">@messages(error.message, error.args:_*)</a></li>
```
DavidGregory084, line 25: Could you explain what is going on in this bit? 😛

conf/messages
```diff
@@ -818,10 +821,10 @@ case.v2.liability.c592.date_for_repayment=Date for repayment
 
 case.v2.liability.c592.details_edit.information=Information from the C592
 case.v2.liability.c592.details_edit.entry_number=Entry number
-case.v2.liability.c592.details_edit.date_p=Entry date
+case.v2.liability.c592.details_edit.date_p=Liability entry date
 case.v2.liability.c592.details_edit.date_hint=Date recorded on the C592 by compliance or port officer, for example, 27 3 2015.
 case.v2.liability.c592.details_edit.bti_reference=BTI reference
-case.v2.liability.c592.details_edit.bti_reference_error=Enter a correct BTI reference number
+case.v2.liability.c592.details_edit.bti_reference_error=Enter a BTI/ ATaR case reference
```
DavidGregory084, line 827: Really small thing, the spacing on BTI/ATaR is not the same between the two examples here

DavidGregory084 (approved):
Looks good to me! 👍

kuldeepwadhwa9 (approved):

app/views/partials/error_summary.scala.html
oliwiaKot: We need the href in the error messages to point to the correct input by the input's id. The id for an input form is being constructed in the component with '_' e.g. 'date_day' but the error key for it would be 'day.day' so we are replacing the '.' with '_' of the error key in the href

PR #475, opened 2021-01-26
Dit 2678 - models for management reporting

oliwiaKot:
Initial models and forms with mappings created for the reporting filters and options, first unit test written - could do with reviewing the mappings, needs unit tests and adding the models to the backend still


DavidGregory084:
Superseded by #513

PR #480, opened 2021-01-29
Dit 2457

ansahdavid:


app/controllers/v2/MyCasesController.scala
```diff
@@ -41,35 +42,54 @@ class MyCasesController @Inject() (
...
+          events =>
+            val eventsLatestFirst = events.results.sortBy(_.timestamp)(Event.latestFirst)
+            val latestCompletedEvent = eventsLatestFirst.collectFirst {
+              case event @ Event(_, _, _, caseReference, _) => Map(caseReference -> event)
+              case _                                                                 => Map.empty
+            }
+            latestCompletedEvent.getOrElse(Map.empty)
+        }
+      }
+      .map(_.foldLeft(Map.empty[String, Event])(_ ++ _))
```
DavidGregory084, line 79: Oh yeah I forgot - did you try using the `findCompletionEvents` function we added here? No biggie but might improve performance

app/controllers/v2/MyCasesController.scala
```diff
@@ -41,35 +42,54 @@ class MyCasesController @Inject() (
...
 
   def displayMyCases(activeSubNav: SubNavigationTab = AssignedToMeTab): Action[AnyContent] =
     (verify.authenticated andThen verify.mustHave(Permission.VIEW_MY_CASES)).async {
       implicit request: AuthenticatedRequest[AnyContent] =>
         for {
           cases <- casesService.getCasesByAssignee(request.operator, NoPagination())
           referralEventsByCase <- getReferralEvents(cases)
+          completeEventsByCase <- getCompletedEvents(cases)
+          _ = logger.info(completeEventsByCase.toString())
```
DavidGregory084, line 88: Can we remove this logging before merging?

DavidGregory084 (approved):
Looks great 👍

app/controllers/v2/MyCasesController.scala
ansahdavid: Good point David and have now sorted this, thanks.

app/controllers/v2/MyCasesController.scala
ansahdavid: Yep removed thanks

PR #481, opened 2021-01-29
DIT-2640 replaced old keyword autocomplete input with an accessible component

oliwiaKot:


app/views/components/input_accessible_auto_complete.scala.html
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    <div id="keyword-search-wrapper" class="autocomplete-wrapper"></div>
+
+
+</div>
+
+<script src="@routes.Assets.at("javascripts/accessible-autocomplete.min.js")" type="text/javascript"></script>
+
+<script type="text/javascript">
```
shiromi, line 54: Is it possible to move this into js file?

oliwiaKot: Tried it - script couldn't be moved to a js file as it needs the list of keywords passed form the page into the js function

shiromi (approved):

PR #489, opened 2021-02-03
Dit 2599 - notification banners and keywords tab frontend work

oliwiaKot:
Do not merge before DIT-2640 (https://github.com/hmrc/tariff-classification-frontend/pull/481)
This covers tickets DIT-2599 and DIT-2633 - notification banners for adding and removing keywords and taking a case off referral, aligning the keywords tab with the prototype, removing tabIndex properties from the tags in templates.


app/views/partials/keywords_details.scala.html
```diff
@@ -22,27 +22,66 @@
...
+
+            }
+
+            <h3 class="heading-medium">Case keyword list</h3>
+
+            @if(keywords.caseKeywords.size<3){
                 <div>
                     <p>
-                        Cases must have 3 or more keywords.
+                            @if(3 - keywords.caseKeywords.size == 1){
```
periklas, line 70: Since this is a constant, what if we add it in the AppConfig?

periklas (approved):
LGTM 👍

PR #494, opened 2021-02-11
Dit 2691 - manage user view with user details and cases tabs

oliwiaKot:
Implemented the view and queries for "Manage user" screen. Delete button, move cases and change teams not implemented - covered in different tickets.

Waiting for confirmation about the total days for ATaRs and Liabilities - whether we need this or not.


gorkemunsal (approved):

app/models/Operator.scala
```diff
@@ -37,6 +37,9 @@ case class Operator(
   def addPermissions(addedPermissions: Set[Permission]): Operator =
     this.copy(permissions = permissions ++ addedPermissions)
 
+  def getMemberTeamNames: Seq[String] = memberOfTeams.flatMap(teamId => Queues.queueById(teamId).map(_.name))
+
+  def isGateway: Boolean = memberOfTeams.contains("1")
```
DavidGregory084, line 42: Could we use `Queues.gateway.id` here?

app/connector/BindingTariffClassificationConnector.scala
```diff
@@ -183,4 +186,22 @@ class BindingTariffClassificationConnector @Inject() (
...
+    withMetricsTimerAsync("update-user") { _ =>
+      val url = s"${appConfig.bindingTariffClassificationUrl}/users/${o.id}"
+      client.PUT[Operator, Operator](url = url, body = o)
+    }
+
+  def getUserDetails(id: String)(implicit hc: HeaderCarrier): Future[Operator] =
+    withMetricsTimerAsync("update-user") { _ =>
+      val url = s"${appConfig.bindingTariffClassificationUrl}/users/$id"
+      client.GET[Operator](url = url)
+    }
```
DavidGregory084, line 199: Really quick thing - it's possible for this function to be called for an ID that does not exist by force-browsing to `/users/user/999` for example, so it would be better if this function returned `Future[Option[User]]` and the caller presented an error page when it is `None`.

I have updated it in this other PR because there I needed to know whether a user already existed or not when creating users so no need to change it here :) 
https://github.com/hmrc/tariff-classification-frontend/pull/495/files#diff-e61cb87110d42f6bd01268a9d96e0618f296ec385dd98d227a95c2abf11a8367R192
https://github.com/hmrc/tariff-classification-frontend/pull/495/files#diff-3e2d2bf4e14c1df3b447923bcbcd32b040b09cd37ad2dabed08666428278e70fR49

app/models/Operator.scala
```diff
@@ -23,8 +23,8 @@ case class Operator(
   name: Option[String]         = None,
   email: Option[String]        = None,
   role: Role                   = Role.CLASSIFICATION_OFFICER,
-  memberOfTeams: List[String]  = List.empty,
-  managerOfTeams: List[String] = List.empty,
+  memberOfTeams: Seq[String]   = Seq.empty,
+  managerOfTeams: Seq[String]  = Seq.empty, //being a member is the same as being the manager, don't use this
```
DavidGregory084, line 27: Shall we remove this from the backend and frontend in a follow-up PR?

app/views/v2/user_heading.scala.html
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import models.viewmodels.CaseHeaderViewModel
+
+@this(
+        caseStatus: views.html.v2.case_status
```
DavidGregory084, line 20: Should this injected view be removed?

app/views/partials/users/view_user.scala.html
```diff
@@ -0,0 +1,108 @@
...
+@import views.html.includes.main
+@import views.html.components.{manager_tools_secondary_navigation, primary_navigation}
+@import models.viewmodels.{SubNavigationTab, ManagerToolsUsersTab, ManagerToolsTab, ApplicationTabViewModel}
+@import controllers.v2.routes
+
+@this(
+user_heading: views.html.v2.user_heading,
+user_details: views.html.partials.users.user_details,
+case_action_buttons_partial: views.html.partials.liabilities.case_action_buttons_partial,
+advanced_search_cases: views.html.partials.liabilities.advanced_search_cases,
```
DavidGregory084, line 33: Are these `case_action_buttons_partial` and `advanced_search_cases` used?

DavidGregory084 (approved):
This looks really good, just a couple comments 💯

app/models/Operator.scala
oliwiaKot: yes, that would be good

app/connector/BindingTariffClassificationConnector.scala
oliwiaKot: great, thanks!

PR #496, opened 2021-02-12
Dit 2695 - delete user journey

oliwiaKot:
Do not merge before DIT-2691
Created views for a confirmation page and the "cases must be moved" page and a controller endpoint to post the confirmation to and delete the user.

### DIFF
--- a/app/controllers/v2/ManageUserController.scala
+++ b/app/controllers/v2/ManageUserController.scala
@@ -34,6 +34,7 @@ import models.forms.v2.RemoveUserForm
 import play.api.data.Form
 import javax.inject.{Inject}
 import scala.concurrent.Future.successful
+import scala.concurrent.Future
 
 class ManageUserController @Inject() (
   verify: RequestActions,
--- a/app/controllers/v2/ManageUserController.scala
+++ b/app/controllers/v2/ManageUserController.scala
@@ -77,15 +77,14 @@ class ManageUserController @Inject() (
           userCases <- casesService.getCasesByAssignee(Operator(pid), NoPagination())
           user      <- userService.getUser(pid)
         } yield {
-          user.isDefined match {
-            case true => {
-              if (userCases.results.nonEmpty) {
-                Ok(cannotDeleteUser(user.get))
-              } else {
-                Ok(confirmDeleteUser(user.get, removeUserForm))
-              }
+          if (user.isDefined) {
+            (user.get.id, userCases.results.nonEmpty) match {
+              case (request.operator.id, _) => Redirect(controllers.routes.SecurityController.unauthorized())
+              case (_, true)                => Ok(cannotDeleteUser(user.get))
+              case (_, _)                   => Ok(confirmDeleteUser(user.get, removeUserForm))
             }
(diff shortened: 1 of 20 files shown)


app/service/UserService.scala
```diff
@@ -43,5 +43,12 @@ class UserService @Inject() (
...
       userDetails <- connector.getUserDetails(id)
-    } yield userDetails
+    } yield {
+      if (userDetails.deleted) {
+        throw new Exception("User with pid " + id + " is deleted")
+      } else {
+        userDetails
+      }
+    }
+
```
DavidGregory084, line 53: Following up from my comment on the other PR, I think this should return `Option` and filter out `deleted` users as `None`, i.e.

```scala
connector
  .getUserDetails(id) // The connector method returns Future[Option[Operator]]
  .map(_.filterNot(_.deleted)) // Filter out deleted users in the inner Option
```

app/controllers/v2/ManageUserController.scala
```diff
@@ -56,4 +60,47 @@ class ManageUserController @Inject() (
...
+            for {
+              user <- userService.getUser(pid)
+            } yield Ok(confirmDeleteUser(user, errors)), {
+            case true =>
+              for {
+                user <- userService.getUser(pid)
+              } yield {
+                val updatedUser = user.copy(deleted = true)
+                userService
+                  .updateUser(updatedUser, request.operator)
```
DavidGregory084, line 95: Did we add a markDeleted endpoint in the backend? Ignore me if not

DavidGregory084 (approved):
Looks great 👍  just got some more suggestions relating to the `getUserDetails` API

app/service/UserService.scala
oliwiaKot: yeah you're right, done this - thanks a lot!

app/controllers/v2/ManageUserController.scala
oliwiaKot: We did, my bad I didn't notice this, I added a service and connector method to use markDeleted and call that in the controller now

app/service/UserService.scala
```diff
@@ -41,5 +41,15 @@ class UserService @Inject() (
...
+    for {
+      userDetails <- connector.getUserDetails(id).map(_.filterNot(_.deleted))
+    } yield userDetails
+
+  def markDeleted(user: Operator, operatorMakingTheChange: Operator)(
+    implicit hc: HeaderCarrier
+  ): Future[Operator] =
+    for {
+      deleted <- connector.markDeleted(user)
+      _ = auditService.auditUserUpdated(user, operatorMakingTheChange)
```
DavidGregory084, line 53: I think this should provide `deleted` to the audit function

DavidGregory084: Ah I see we only use the `id` of the operator in the audit payload so this probably doesn't matter

DavidGregory084 (approved):
👍

app/service/UserService.scala
oliwiaKot: yeah it uses the id, but I changed that anyway, good spot!

vajarova (approved):

PR #498, opened 2021-02-12
DIT 2692 - Change team view and form

emilia-trendafilova:


app/controllers/v2/ManageUserController.scala
```diff
@@ -0,0 +1,118 @@
...
+)(
+  implicit val appConfig: AppConfig,
+  ec: ExecutionContext
+) extends FrontendController(mcc)
+    with I18nSupport
+    with Logging {
+
+  private val userEditTeamForm = UserEditTeamForm.editTeamsForm
+
+  private def getReferralEvents(
```
oliwiaKot, line 53: might need getting up to date with DIT-2692 as the methods to get referral events were not needed and got removed

oliwiaKot (approved):
lgtm

app/controllers/v2/ManageUserController.scala
periklas: Perfect, thanks! We have removed these two methods 👍

PR #502, opened 2021-02-19
DIT-2754 | view for manage keywords 

warrenskiwork1979:


kuldeepwadhwa9 (approved):

test/unit/views/managementtools/KeywordsTabSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    "include an edit keyword button" in {
+      val doc = view(
+        manageKeywordsView()
+      )
+      doc.getElementById("keyword_search-edit-keyword-button") should containText("Edit keyword")
+    }
+
+    //TODO Test search box
```
emilia-trendafilova, line 67: is this needed?

emilia-trendafilova (approved):
LGTM

app/models/viewmodels/managementtools/Keyword.scala
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.viewmodels.managementtools
+
+import models.CaseStatus.CaseStatus
+import models.{ApplicationType, CaseStatus}
+
+case class Keyword(
```
emilia-trendafilova, line 22: Can you check that the model for Keyword matches the one in the backend? For reference : DIT-2760

emilia-trendafilova (approved):

app/models/viewmodels/managementtools/Keyword.scala
warrenskiwork1979: So this is just a stub at the moment for this ticket which will be replaced with the new Keyword model matching the backend in DIT-2755, which is in progress.

PR #539, opened 2021-03-09
Dit 2757

ansahdavid:


app/controllers/v2/ManageKeywordsController.scala
```diff
@@ -84,27 +84,77 @@ class ManageKeywordsController @Inject()(
 
   def createKeyword(activeSubNav: SubNavigationTab = ManagerToolsKeywordsTab): Action[AnyContent] =
     (verify.authenticated andThen verify.mustHave(Permission.MANAGE_USERS)).async { implicit request =>
-
-      keywordService.findAll(NoPagination()).flatMap {
-        keywords =>
-          val keywordNames = keywords.results.map(_.name)
-          KeywordForm.formWithAuto(keywordNames).bindFromRequest.fold(
+      keywordService.findAll(NoPagination()).flatMap { keywords =>
```
DavidGregory084, line 87: I think you might need to pull from master and update this as this method doesn't take a `Pagination` argument anymore

app/controllers/v2/ManageKeywordsController.scala
```diff
@@ -84,27 +84,77 @@ class ManageKeywordsController @Inject()(
...
+                         Redirect(
+                           controllers.v2.routes.ManageKeywordsController
+                             .displayKeywordChangeConfirmation(
+                               savedKeyword.name,
+                               savedKeyword.approved,
+                               c.application.goodsName
+                          )
+                      )
+                    }
+                  case "REJECT" =>
```
DavidGregory084, line 139: It might be nice to use a model here like `ChangeKeywordStatusAction` or something like that rather than a String but it's not urgent

app/models/KeywordStatus.scala
```diff
@@ -0,0 +1,29 @@
...
+  type KeywordStatus = Value
+  val APPROVED, REJECTED, REPLACED = Value
+
+  def format(status: KeywordStatus): String =
+    status match {
+      case APPROVED => "Approved"
+      case REJECTED => "Rejected"
+      case REPLACED => "Renamed"
+    }
+}
```
DavidGregory084, line 29: Is this model used anywhere?

DavidGregory084 (approved):
Looks good 👍

app/controllers/v2/ManageKeywordsController.scala
ansahdavid: Model created and now in use

app/models/KeywordStatus.scala
ansahdavid: Newer version this model has now been created and is in use

PR #554, opened 2021-03-12
DIT-2799 added correspondence and misc filters, added case source and case des…

oliwiaKot:
…cription fields, fixed completed status filter


DavidGregory084 (approved):
💯

PR #558, opened 2021-03-12
DIT-2815 frontend changes to edit liability ruling

oliwiaKot:


app/views/v2/edit_liability_ruling.scala.html
```diff
@@ -75,10 +80,12 @@
 
   @confirm_or_cancel_buttons(
    id = "edit_liability_decision",
-   confirmText = messages("case.v2.liability.save_and_continue_button"),
-   cancelText = messages("case.v2.liability.cancel_button"),
+   confirmText = "Save changes",
```
vajarova, line 83: should we aim to keep all messages in the message files, especially for when it comes to translating them

vajarova (changes requested):

app/views/v2/edit_liability_ruling.scala.html
oliwiaKot: I added messages for the confirm text and cancel text

vajarova: Thanks @oliwiaKot

vajarova (approved):

PR #567, opened 2021-03-16
Dit 2828

ansahdavid:
Updated functionality of the Back link displayed in web pages


test/unit/views/partials/BackLinkViewSpec.scala
```diff
@@ -55,128 +46,74 @@ class BackLinkViewSpec extends ViewSpec {
...
+      val doc = view(back_link()(messages))
 
       // Then
       doc                             should containElementWithID("back-link")
       doc.getElementById("back-link") should haveTag("a")
-      doc.getElementById("back-link") should haveAttribute("href", "url")
-      doc.text()                      shouldBe "Back to " + customLabel
+      doc.getElementById("back-link") should haveAttribute("href", "#")
+      doc.text()                      shouldBe "Back"
     }
```
DavidGregory084, line 117: I think all of these the tests in this spec are pretty much the same now, so perhaps we could get rid of all but one?

DavidGregory084 (approved):
Looks good 👍

test/unit/views/partials/BackLinkViewSpec.scala
ansahdavid: Yep good point David, I'll update them now.

ansahdavid: Done

PR #569, opened 2021-03-17
DIT-2357 Upscan integration

DavidGregory084:
This changes the file upload forms on the attachments tab to use Upscan and splits each case status change page into two pages, one to enter the reason for the status change and one to upload a supporting email using Upscan.


kuldeepwadhwa9 (approved):

ansahdavid (approved):
Looks good

PR #571, opened 2021-03-17
DIT-2357 preserve single-page journey using jquery

DavidGregory084:
Don't merge - this is a proof of concept


DavidGregory084:
Overall thoughts
* this is really gross
* it would require lots of reverse engineering and messing around in main.js to reproduce the existing form error behaviours
* it's not testable
* the upscan-stub doesn't produce realistic responses for JSON requests so it's even less testable

PR #573, opened 2021-03-18
DIT-2880 made the max date inclusive

oliwiaKot:


app/models/forms/ReportDateForm.scala
```diff
@@ -41,6 +41,13 @@ object ReportDateForm {
       .toInstant
   }
 
+  private val formDate2InstantAtEndOfDay: DateForm => Instant = { dateForm =>
+    LocalDate
+      .of(dateForm.year.toInt, dateForm.month.toInt, dateForm.day.toInt + 1)
+      .atStartOfDay(ZoneOffset.UTC)
+      .toInstant
```
DavidGregory084, line 48: I think we need to be careful here - `dateForm.day.toInt + 1` could produce an invalid date even if the user enters something valid, e.g. if they entered the 31st of a month. How about using `plusDays` instead:

```suggestion
    LocalDate
      .of(dateForm.year.toInt, dateForm.month.toInt, dateForm.day.toInt)
      .plusDays(1)
      .atStartOfDay(ZoneOffset.UTC)
      .toInstant
```

DavidGregory084 (changes requested):
Looks good, just a comment about how we are incrementing the date 👍

app/models/forms/ReportDateForm.scala
oliwiaKot: thank you! I've applied that now

DavidGregory084 (approved):
👍

vajarova (approved):

PR #578, opened 2021-03-18
DIT-2783 Correspondence case header

shiromi:


periklas (approved):

ansahdavid (approved):

conf/messages
```diff
@@ -941,7 +941,7 @@ case.v2.liability.reopen.button=Reopen case
 
 case.v2.liability.advanced_search.button=Search cases
 
-case.v2.correspondence.page_title=Correspondence case details
```
vajarova, line 944: can we keep those (misc and corrie and liability) the same as per Eszter's comment on the ticket please

vajarova (changes requested):

vajarova (approved):

PR #582, opened 2021-03-19
DIT-2802 Reject case page and DIT-2804 Suspend ATaR & liability case page

shiromi:
DIT-2802
DIT-2804


emilia-trendafilova (approved):
LGTM

PR #598, opened 2021-03-25
Dit 2897 - Add audit events for case update

emilia-trendafilova:


app/audit/AuditService.scala
```diff
@@ -129,6 +129,14 @@ class AuditService @Inject() (auditConnector: DefaultAuditConnector) {
       auditPayload   = baseAuditPayload(c, operator) + ("comment" -> "Liability case created")
     )
 
+  def auditCaseUpdated(originalCase: Case, updatedCase: Case, operator: Operator)(implicit hc: HeaderCarrier): Unit =
+    sendExplicitAuditEvent(
+      auditEventType = CaseUpdated,
+      auditPayload = baseAuditPayload(originalCase, operator) + (
+        "updatedCase" -> updatedCase.toString
```
DavidGregory084, line 136: I think this should include the full `originalCase` and `updatedCase` JSON so that it's possible to see what has changed

app/controllers/v2/MoveCasesController.scala
```diff
@@ -408,7 +408,7 @@ class MoveCasesController @Inject() (
       .flatten
       .map(c =>
         for {
-          updatedCase <- casesService.updateCase(c.copy(assignee = user, queueId = Some(teamId)))
+          updatedCase <- casesService.updateCase(c.copy(assignee = user, queueId = Some(teamId)), user.getOrElse(Operator(originalUserId)))
```
DavidGregory084, line 411: I think this should be `request.operator` as the second argument is the `operatorUpdating`. At the moment it will use the new assignee for the case as the person that did the update I think?

app/service/CasesService.scala
```diff
@@ -457,8 +460,11 @@ class CasesService @Inject() (
   def getAssignedCases(pagination: Pagination)(implicit hc: HeaderCarrier): Future[Paged[Case]] =
     connector.findAssignedCases(pagination)
 
-  def updateCase(caseToUpdate: Case)(implicit hc: HeaderCarrier): Future[Case] =
-    connector.updateCase(caseToUpdate)
+  def updateCase(caseToUpdate: Case, operator: Operator)(implicit hc: HeaderCarrier): Future[Case] =
+    for {
+      updatedCase <- connector.updateCase(caseToUpdate)
+      _ = auditService.auditCaseUpdated(caseToUpdate, updatedCase, operator)
```
DavidGregory084, line 466: This does not include the original case, it will add the same case twice to the audit event I think?

I think we need to add an extra argument to this function that takes the case before it was updated.

DavidGregory084 (changes requested):
I think we need to be a little bit careful about what data we are using here

In an audit event we want to see:
1. What the data was before we did any changes to it
2. What the data has been updated to now
3. Who updated it - this can be different from the assignee for the case

I think there are a few occasions here where we are not showing 1. and 3. - let's have a look at this together tomorrow?

app/audit/AuditService.scala
emilia-trendafilova: Thanks. Resolved

app/controllers/v2/MoveCasesController.scala
emilia-trendafilova: Thanks. Resolved

DavidGregory084 (approved):
Looks great 💯

PR #612, opened 2021-04-12
DDCE-1630: ATAR handover file upload issue

abhijeetardale:


app/connector/FileStoreConnector.scala
```diff
@@ -79,6 +82,9 @@ class FileStoreConnector @Inject() (
 
   def initiate(request: FileStoreInitiateRequest)(implicit hc: HeaderCarrier): Future[FileStoreInitiateResponse] =
     withMetricsTimerAsync("initiate-file-upload") { _ =>
+
+      Logger.error(s"\n\n initiate-file-upload :" + request + "\n\n")
```
DavidGregory084, line 86: `play.api.Logger` is deprecated, but you can replace it by importing `play.api.Logging` and extending `Logging` to get a `logger` member in this class

app/controllers/PdfDownloadController.scala
```diff
@@ -48,9 +48,9 @@ class PdfDownloadController @Inject() (
               url     <- OptionT.fromOption[Future](meta.url)
               content <- OptionT(fileStore.downloadFile(url))
             } yield Ok
-              .streamed(content, None, Some(meta.mimeType))
+              .streamed(content, None, meta.mimeType)
               .withHeaders(
-                "Content-Disposition" -> s"attachment; filename=${meta.fileName}"
+                "Content-Disposition" -> s"attachment; filename=${meta.fileName.getOrElse("New Attachment")}"
               )
```
DavidGregory084, line 54: This determines the actual filename that will be downloaded - I think it might be better to do something like this:

```scala
           val pdfResult = for {
              pdf      <- OptionT.fromOption[Future](decision.decisionPdf)
              meta     <- OptionT(fileStore.getFileMetadata(pdf.id))
              url      <- OptionT.fromOption[Future](meta.url)
              fileName <- OptionT.fromOption[Future](meta.fileName)
              content  <- OptionT(fileStore.downloadFile(url))
            } yield Ok
              .streamed(content, None, Some(meta.mimeType))
              .streamed(content, None, meta.mimeType)
              .withHeaders(
                "Content-Disposition" -> s"attachment; filename=${fileName}"
              )
```
Note the additional generator in the for comprehension - this is because Upscan must have sent us a callback with the file info by the point that we are able to download a file

app/controllers/PdfDownloadController.scala
```diff
@@ -77,9 +77,9 @@ class PdfDownloadController @Inject() (
               url     <- OptionT.fromOption[Future](meta.url)
               content <- OptionT(fileStore.downloadFile(url))
             } yield Ok
-              .streamed(content, None, Some(meta.mimeType))
+              .streamed(content, None, meta.mimeType)
               .withHeaders(
-                "Content-Disposition" -> s"attachment; filename=${meta.fileName}"
+                "Content-Disposition" -> s"attachment; filename=${meta.fileName.getOrElse("New Attachment")}"
```
DavidGregory084, line 82: Same suggestion here

app/controllers/PdfDownloadController.scala
```diff
@@ -104,9 +104,9 @@ class PdfDownloadController @Inject() (
           url     <- OptionT.fromOption[Future](meta.url)
           content <- OptionT(fileStore.downloadFile(url))
         } yield Ok
-          .streamed(content, None, Some(meta.mimeType))
+          .streamed(content, None, meta.mimeType)
           .withHeaders(
-            "Content-Disposition" -> s"attachment; filename=${meta.fileName}"
+            "Content-Disposition" -> s"attachment; filename=${meta.fileName.getOrElse("New Attachment")}"
```
DavidGregory084, line 109: Same suggestion here

app/controllers/v2/AtarController.scala
```diff
@@ -64,6 +65,9 @@ class AtarController @Inject() (
     keywordForm: Form[String]            = KeywordForm.form,
     uploadForm: Form[String]             = UploadAttachmentForm.form
   )(implicit request: AuthenticatedCaseRequest[_]): Future[Html] = {
+
+    Logger.error(s"\n\n fileId :" + fileId + "\n\n")
```
DavidGregory084, line 69: Same comment about `play.api.Logger` as above

app/service/FileStoreService.scala
```diff
@@ -106,7 +108,7 @@ class FileStoreService @Inject() (connector: FileStoreConnector) extends Logging
     connector.delete(fileId)
 
   private def toFileAttachment(size: Long): FileMetadata => FileStoreAttachment = { r =>
-    FileStoreAttachment(r.id, r.fileName, r.mimeType, size)
+    FileStoreAttachment(r.id, r.fileName.getOrElse(""), r.mimeType.getOrElse(""), size)
   }
```
DavidGregory084, line 112: You might be able to get rid of this method entirely as I don't think the `upload` method above is used anymore

app/models/StoredAttachment.scala
```diff
@@ -73,8 +73,8 @@ object StoredAttachment {
       operator    = attachment.operator,
       timestamp   = attachment.timestamp,
       url         = metadata.url,
-      fileName    = metadata.fileName,
-      mimeType    = metadata.mimeType,
+      fileName    = metadata.fileName.getOrElse("New Attachment"),
+      mimeType    = metadata.mimeType.getOrElse("FileType"),
```
DavidGregory084, line 77: It might be better to do this in the places where the `fileName` and `mimeType` are used rather than here as this is dummy data for display purposes only, so perhaps it belongs in the Twirl templates where it is rendered?

app/models/request/FileStoreInitiateRequest.scala
```diff
@@ -19,6 +19,8 @@ package models.request
 import play.api.libs.json.{ OFormat, Json }
 
 case class FileStoreInitiateRequest(
+  fileName: Option[String] = None,
+  mimeType: Option[String] = None,
```
DavidGregory084, line 23: I don't think that this change is needed - the filestore will ignore it - see here https://github.com/hmrc/binding-tariff-filestore/blob/master/app/uk/gov/hmrc/bindingtarifffilestore/model/upscan/v2/FileStoreInitiateRequest.scala

app/models/StoredAttachment.scala
DavidGregory084: To explain this a bit more, we should not be able to download a file when we do not know the mime type.

In the binding-tariff-filestore, the [SuccessfulScanResult](https://github.com/hmrc/binding-tariff-filestore/blob/master/app/uk/gov/hmrc/bindingtarifffilestore/model/upscan/ScanResult.scala#L27) contains the [UploadDetails](https://github.com/hmrc/binding-tariff-filestore/blob/master/app/uk/gov/hmrc/bindingtarifffilestore/model/upscan/ScanResult.scala#L57). 

Upscan always returns this info, so this means that if our file was successfully virus scanned, we must know its name and mime type.

app/controllers/ViewAttachmentController.scala
```diff
@@ -47,7 +47,7 @@ class ViewAttachmentController @Inject() (
             url     <- OptionT.fromOption[Future](fileSubmitted.url)
             content <- OptionT(fileService.downloadFile(url))
```
DavidGregory084, line 48: You could potentially add a `mimeType` generator here too to return 404 rather than an empty mime type? I don't think this one makes a huge difference either way 😛

app/models/StoredAttachment.scala
DavidGregory084: This means that if we do not know the file's name or mime type, it must either have failed its virus scan or the scan has not completed yet

DavidGregory084 (approved):
This looks good, nice work! Just some suggestions which you can apply if you feel like it 💯

DavidGregory084 (approved):
Looks great 👍

PR #616, opened 2021-05-21
[DDCE-1906]:Fix problem opening attachments in ATaR operational service

anburoach:


app/views/partials/attachments_edit_list.scala.html
```diff
@@ -32,14 +32,14 @@
...
-        <th scope="col" lass="bold-small w-25 p-0" id="header-select" >Publish</th>
+        <th scope="col" class="bold-small w-25 p-0" id="header-select" >Publish</th>
         <th scope="col" class="bold-small p-0 attachment-table-date">Date</th>
     </tr>
     @for((file: StoredAttachment, index: Int) <- attachments.zipWithIndex) {
     <tr id="@{id}-row-@{index}" class="border-bottom row-alignment">
         <td>
             <div id="@{id}-row-@{index}-title" class="file-name-column-width w-75 pl-0">
-                @attachment(s"$id-row-$index", file, useTabIndex = startAtTabIndex.map(x => x+1+(0*index)))
+                @attachment(s"$id-row-$index", file, caseRef.getOrElse(""), useTabIndex = startAtTabIndex.map(x => x+1+(0*index)))
```
DavidGregory084, line 42: I would be really cautious about using this `getOrElse` here. I think if you do that you will want to make sure that the `caseRef` above does not have a default value to ensure that it is passed in. It might even be better to make it non-optional in the constructor for this template?

app/views/partials/attachments_list.scala.html
```diff
@@ -45,7 +45,7 @@
                     @for((file: StoredAttachment, index: Int) <- attachments.zipWithIndex) {
                         <tr id="@{id}-row-@{index}">
                             <td id="@{id}-row-@{index}-title" >
-                                @attachment(s"$id-row-$index", file, useTabIndex = startAtTabIndex.map(x => x+1+(0*index)))
+                                @attachment(s"$id-row-$index", file, caseRef.getOrElse(""), useTabIndex = startAtTabIndex.map(x => x+1+(0*index)))
```
DavidGregory084, line 48: Same comment here

DavidGregory084 (changes requested):
Looks good, would just prefer to avoid using that `getOrElse` as I think it will cause a different kind of problem opening files

PR #617, opened 2021-06-16
Ddce 1975

PasGat:


yangwong9664 (changes requested):
- Some dependencies can be updated to a more recent version of play27: play-json-union-formatter, play-ui, service-integration-test, simple-reactivemongo, couple others too, you can run `sbt dependencyUpdates` to see them
- There's quite a few warnings, if you run `sbt clean compile` you should see them, they don't all have to be fixed, just the simple easy to fix ones like Deprecated classes, unused imports etc.

app/filters/Filters.scala
```diff
@@ -18,9 +18,7 @@ package filters
...
 import com.google.inject.Inject
 import play.api.http.DefaultHttpFilters
-import uk.gov.hmrc.play.bootstrap.frontend.filters.FrontendFilters
 
 class Filters @Inject() (
-  shutterFilter: ShutteringFilter,
-  frontendFilters: FrontendFilters
-) extends DefaultHttpFilters(frontendFilters.filters :+ shutterFilter: _*)
+  shutterFilter: ShutteringFilter
+) extends DefaultHttpFilters(shutterFilter)
```
DavidGregory084: FWIW I don't think you will want to remove these `FrontendFilters` completely here, or if you do you need to enable them another way.

I think when this code was originally written in 2018 or so this was the way to set up these filters but it seems like the advice has changed according to the [bootstrap-play README](https://github.com/hmrc/bootstrap-play#filters).

So it seems like if you remove the FrontendFilters here you will probably also need to remove the [line which sets this as the default filters](https://github.com/hmrc/tariff-classification-frontend/blob/master/conf/application.conf#L37), delete this file entirely and then add the following line to `application.conf`:
```scala
play.filters.enabled += "filters.ShutteringFilter"
```
Hope that helps

PR #645, opened 2023-01-06
DDCE-3642: remove all references to covid from cover letter

rreddy80:
Remove the sentence of "However, if you need more time due to coronavirus, you can be given an extra 3 months if the extension is asked for within these 30 days."  & "Due to coronavirus (COVID-19), you will be able to ask for an extra 3 months for your appeals. This is for both an HMRC review and the independent tribunal. " from all ATaR letters.


mywyau (approved):
Passing tests and LGTM

PR #652, opened 2023-03-29
DDCE-3845 - handle attachment non Ascii characters

frikit:


app/assets/javascripts/tab_focus.js
```diff
@@ -48,5 +48,12 @@ function fixAnchorInURL(targetURL, csrf) {
     if (anchor.toString() === "") {
         //no anchor in url
         getAnchor(targetURL, csrf)
+    } else if (anchor.includes("&")) {
```
frikit, line 51: there were a bug when upload file with more than 10mb the error returned back from upscan had wrong url, the anchor was in the middle when it supposed to be at the end, fixed that here

scalastyle-config.xml
```diff
@@ -42,7 +42,7 @@
             <parameter name="maxParameters"><![CDATA[8]]></parameter>
         </parameters>
     </check>
-    <check level="warning" class="org.scalastyle.scalariform.MagicNumberChecker" enabled="true">
+    <check level="warning" class="org.scalastyle.scalariform.MagicNumberChecker" enabled="false">
```
frikit, line 45: disabled this rule for now as its limited to 100 warnings and all 100 are magic numbers, need to cleanup some before enabling it back

app/controllers/ViewAttachmentController.scala
```diff
@@ -43,6 +45,16 @@ class ViewAttachmentController @Inject() (
...
     with WithUnsafeDefaultFormBinding {
 
+  /*
+   * This method is used to encode the filename in the Content-Disposition header
+   * url encode the filename and browser will decoded based on the charset you specify
+   * if no filename is provided, generate a random one
+   */
+  private def getEncodedFilenameHeader(fileSubmitted: FileMetadata): String =
+    fileSubmitted.fileName
+      .map(fileName => s"filename*=UTF-8''${URLEncoder.encode(fileName, "UTF-8")}")
```
EdCope, line 55: I did see some examples where they would keep the `filename=` as well as `filename*=` as a fallback for older browser support. Should we expect to maintain older support like that ?

Example: [Mozilla Content-Disposition - Compatibility notes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#compatibility_notes)

app/views/includes/govuk_wrapper.scala.html
```diff
@@ -57,7 +57,7 @@
 
 @govukLayout(
     pageTitle = Some(pageTitle),
-    headBlock = Some(hmrcInternalHead(head, nonce = CSPNonce.get)),
+    headBlock = Some(hmrcInternalHead(head)),
```
EdCope, line 60: Remove CSPNonce import too

app/controllers/ViewAttachmentController.scala
frikit: this is internal app when I was part of ATAR team the gov staff were moving slowing to edge and edge is chrome basically so I dont expect any problems at all here, and also if you look in notes there on mozila you see firefix handles this since v5 so its been over a decade

EdCope: Makes sense thanks

EdCope (approved):

PR #654, opened 2023-04-06
DDCE-3288 - Open case details page when accessing a case without ruli…

frikit:
…ng tab


test/unit/models/response/UploadErrorSpec.scala
```diff
@@ -14,17 +14,27 @@
  * limitations under the License.
  */
 
-package views
-
-object CaseDetailPage extends Enumeration {
```
frikit, line 19: this is bug in git,. I just deleted one file and created test for another class

EdCope (approved):

PR #656, opened 2023-04-24
DDCE-3736 Update Content Secure Policy 

mywyau:
**Changes**
- Adding nonces
- Updating config
- Removing comments in tests

**So initial issues** 
Changing the csp results in having to add in nonces for script elements there is javascript on the checkboxes that break as they are being blocked from submitting the page when checked. 

Screenshot 2023-04-13 at 14 04 28

Screenshot 2023-04-13 at 14 06 41

This results in breaking the ATs as the checkboxes lose their case filtering capabilities.


mywyau:
**As an update:** 
Refactored the javascript to be out of the component and moved into the relevant view. The two blocked was the remove key word link and checkbox filters on advanced search page as shown above. Using event listener instead and adding a nonce to the script tag to comply with new CSP.

EdCope (changes requested):
CSP was blocking backlinks.
Print inline script also affected under reports sections
File uploads working differently on this one so are ok

app/models/CaseReferral.scala
```diff
@@ -21,4 +21,11 @@ case class CaseReferral(
   reasons: List[ReferralReason.Value],
   note: String,
   referManually: Option[String]
-)
+) {
+
+  val function1 = (x: Int) => x + 1
```
EdCope, line 26: Remove here

EdCope (approved):

PR #661, opened 2023-07-06
DDCE-4386 - Upgrade mongo 0.74.x to 1.3.x

frikit:


app/repositories/SessionRepository.scala
```diff
@@ -21,26 +21,26 @@ import org.mongodb.scala.model.Filters.equal
...
 
 case class DatedCacheMap(
   id: String,
   data: Map[String, JsValue],
-  lastUpdated: LocalDateTime = LocalDateTime.now(ZoneOffset.UTC)
+  lastUpdated: Instant = Instant.now()
 )
 
 object DatedCacheMap {
-  implicit val dateFormat: Format[LocalDateTime] = MongoJavatimeFormats.localDateTimeFormat
```
frikit, line 42: hmrc-mongo has removed LocalDateTime support and I decided to migrate over to instant from implementation of format method doesnt seems to be any difference so should be safe

PR #673, opened 2024-03-15
DDCE-5407 Cases due to expire report

gastonracui:


app/models/reporting/Report.scala
```diff
@@ -216,6 +219,30 @@ object Report {
     statuses  = Set(PseudoCaseStatus.CANCELLED)
   )
 
+  private val calculateDueToExpireDateRange: InstantRange = {
+    val now = LocalDate.now()
+    InstantRange(
+      min = now.minusMonths(1).`with`(firstDayOfMonth()).atStartOfDay(ZoneOffset.UTC).toInstant,
+      max = now.`with`(lastDayOfMonth()).atStartOfDay(ZoneOffset.UTC).toInstant
```
lmugnai, line 226: Can I suggest separating this a bit, just to make it more readable? Something like:
val firstDay = now.minusMonths(1).`with`(firstDayOfMonth())
  val lastDay = now.`with`(lastDayOfMonth())
  InstantRange(
    min = firstDay.atStartOfDay(ZoneOffset.UTC).toInstant,
    max = lastDay.atStartOfDay(ZoneOffset.UTC).toInstant
  )

lmugnai (approved):

PR #687, opened 2024-08-23
DDCE-6003 Try to fix compilation warnings

mywyau:
Not really sure about this ticket, lots of match cases non exhaustive warnings but my opinion is there is often not really suitable replacements for catch all cases.

Regarding implicits we have a ticket for migrating to Scala 3 for this frontend. The code base is large and changing implicits to new Scala 3 syntax is not easy and a lot of changes. Should be done in another ticket. 

https://jira.tools.tax.service.gov.uk/browse/DDCE-5955

This ticket https://jira.tools.tax.service.gov.uk/browse/DDCE-6003 says there is 100 warnings after the spike and Victor's work it had been reduced to only around 35 warnings. 

Most match case errors have no good catch all solutions:

- @unchecked - using these tags to to ignore the error, not a good idea imo
-  Suppress Warnings with Compiler Flags using `scalacOptions ++= Seq("-P:silencer:globalFilters=match may not be exhaustive")`
- add catch alls with case _ => ???
```
status match {
  case LIVE     => "Live"
  case NON_LIVE => "Non-live"
  case _        => "Unknown" 
}



```


conf/messages
```diff
@@ -1325,6 +1325,7 @@ users.user.details.status=Status
 
 users.user.details.role.hint.classification=Manage ATaR, liability, correspondence and miscellaneous cases.
 users.user.details.role.hint.gateway=Assign ATaR cases. Create and assign Liability, Correspondence and Miscellaneous cases.
+users.user.details.role.hint.not.found=Role or isGateway not found
```
mywyau, line 1328: will probably need better and new content for this message and welsh yeah if it ever gets decided we need to handle these match cases or not

app/models/ReferralReason.scala
```diff
@@ -21,8 +21,10 @@ object ReferralReason extends Enumeration {
 
   val REQUEST_SAMPLE, REQUEST_MORE_INFO = Value
 
-  def format(reason: ReferralReason): String = reason match {
+  def format(reason: ReferralReason): String = 
+    reason match {
     case REQUEST_SAMPLE    => "To request a sample"
     case REQUEST_MORE_INFO => "To request more information"
+    case _ => "Unknown"
```
mywyau, line 28: hard coded match cases again how do we handle this?

Taiwo-AB:
This is a challenging situation because there are various methods to address the warnings, such as using sealed traits, unchecked, wildcards etc. I believe a discussion is necessary to determine the best strategy. However, if you choose to use the wildcard, make sure to use proper naming. For now, using unchecked might be the simplest solution until we find a more suitable approach, as we haven’t encountered any issues with this service so far.

PR #694, opened 2024-09-27
Ddce 6060 file upload

palanivel-subramanian:


app/controllers/PdfDownloadController.scala
```diff
@@ -58,13 +58,14 @@ class PdfDownloadController @Inject() (
             val documentType = messages("errors.document-not-found.ruling-certificate")
 
             pdfResult.getOrElseF {
-              caseService.regenerateDocuments(cse, request.operator).flatMap { regeneratedCase =>
-                logger.info(s"[PdfDownloadController][getRulingPdf] new decisionPdf: ${regeneratedCase.decision.flatMap(_.decisionPdf).map(_.id)}")
-                downloadFile(regeneratedCase.decision.flatMap(_.decisionPdf))
+              for {
+                regeneratedCase <- caseService.regenerateDocuments(cse, request.operator)
+                _ = logger.info(s"[PdfDownloadController][getRulingPdf] new decisionPdf: ${regeneratedCase.decision.flatMap(_.decisionPdf).map(_.id)}")
+                fileDownloadResult <- downloadFile(regeneratedCase.decision.flatMap(_.decisionPdf))
```
lmugnai, line 64: when working with for comprehension you don't need to use the getOrElseF inside. The getOrElseF returns a Future[Result], which is fine but I think it would be cleaner if we flatten into a single for comprehension. SOmething like this:
pdfResult.getOrElseF {
  for {
    regeneratedCase <- caseService.regenerateDocuments(cse, request.operator)
    _ = logger.info(s"[PdfDownloadController][getRulingPdf] new decisionPdf: ${regeneratedCase.decision.flatMap(_.decisionPdf).map(_.id)}")
    
    fileDownloadResult <- downloadFile(regeneratedCase.decision.flatMap(_.decisionPdf))
  } yield fileDownloadResult.getOrElseF {
    Future.successful(NotFound(document_not_found(documentType, reference)))
  }
}

lmugnai (approved):

PR #698, opened 2024-11-06
DDCE-6148 ATaR - Amend Definition List

Divya-Manivannan-hmrc:


app/views/partials/sample/sample_status.scala.html
```diff
@@ -21,8 +21,6 @@
 
 @(sample: SampleTabViewModel, startAtTabIndex : Int = 0)(implicit request: AuthenticatedRequest[_], messages: Messages)
 
-<h3 class="govuk-heading-m govuk-!-margin-top-6" id="sample-status-heading">@messages("case.sample.sample-status-header")</h3>
-
 <div id="sample-status" class="govuk-summary-list__row">
```
gastonracui, line 24: i think you are also missing to create a blank dd when the request does not have the EDIT_SAMPLE permission

gastonracui (changes requested):

app/views/partials/sample/sample_status.scala.html
gastonracui: or you can write a tests verifying is not required

PR #703, opened 2024-11-15
DDCE-6148 - Amend Definition List

Divya-Manivannan-hmrc:


lmugnai:
Hi @Divya-Manivannan-hmrc just for my understanding, how does moving the h3 from sample status to sample details address the issue with the definition list? Thanks

lmugnai:
Also, the message key "case.sample.sample-status-header" is referring to sample status and not sample details so it seems strange moving this.

Divya-Manivannan-hmrc:
Screenshot from 2024-11-05 17-46-43
The issue is that <dl> can only have <dt> and <dl> tags in it. But in this case, it had h3 as a direct child. So, i moved it out.

lmugnai (approved):

PR #706, opened 2024-11-25
DDCE-5782 Remove deprecated PDF_GENERATOR_SERVICE

gastonracui:


lmugnai:
@gastonracui can we change this - https://github.com/hmrc/tariff-classification-frontend/blob/d03729e7cdcce235ef3715212298843e785a0816/app/models/forms/v2/CorrespondenceForm.scala#L17-L17 to use the correct package models.forms.v2? There will be a couple of imports and some tests that need updating too.

lmugnai (approved):

PR #713, opened 2025-02-19
DDCE-6420 Replace play-json-union-formatter Library

Divya-Manivannan-hmrc:


SachaMcCormack:
Do all those dependencies need their versions decreased?

SachaMcCormack:
Missing test class for added Union class: https://github.com/hmrc/play-json-union-formatter/blob/main/src/test/scala/uk/gov/hmrc/play/json/UnionSpec.scala

SachaMcCormack (approved):

PR #717, opened 2025-05-29
DDCE-6662

erickbm6922:
ATaR - binding-tariff-classification: Redesign Data Access Pattern to Eliminate caseKeywords View


app/services/RedirectService.scala
```diff
@@ -42,6 +42,6 @@ class RedirectService @Inject() {
       case ApplicationType.MISCELLANEOUS =>
         Redirect(controllers.v2.routes.MiscellaneousController.displayMiscellaneous(reference, fileId))
           .flashing(request2flash)
-      case _ => BadRequest
+      case null => BadRequest
```
lmugnai, line 45: Why are we only checking for null here instead of all possible scenarios?

erickbm6922: You're absolutely right. I changed this because the compiler warned that case _ => was unreachable except for null, I change the code to a different approach to avoid that warning

lmugnai:
@erickbm6922 can you address the PR bot comments please - some dep updates

lmugnai (approved):

